PR #1, opened 2020-06-08
Bi 3867 initial swagger spec

mwestacott:


spec/schema.json
```diff
@@ -0,0 +1,67 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "description": "An API to support lodging objections to compulsory and voluntary strike offs",
+    "version": "1.0.0",
+    "title": "Companies House - Strike Off Objections API"
+  },
+  "host": "api.companieshouse.gov.uk",
+  "basePath": "/company/{companyNumber}/objections",
```
markpit, line 9: not sure about the basePath "/company/{companyNumber}/objections"
there can be other objections, not just strike off objections
maybe "/company/{companyNumber}/objections/strike-off/request/
I admit my RESTful ness is not great, just wondering how to break it down into resources but leave it open for other objection paths in the future

mwestacott: Yep. Fair enough. We have it in the repo name. Will change.

spec/schema.json
```diff
@@ -0,0 +1,67 @@
...
+  "schemes": [
+    "https"
+  ],
+  "tags": [
+    {
+      "name": "Strike Off Objections"
+    }
+  ],
+  "paths": {
+    "/test": {
```
markpit, line 19: How about instead of /test, we just post to the basePath - /company/{companyNumber}/objections/strike-off/request
This could possibly then be the real 'create' endpoint
POST -> /company/{companyNumber}/objections/strike-off/request
and would eventually return the id of the request created. But for now just return some dummy id.

(Just thinking ahead) - Then we could have a /process endpoint which would also allow support to re-process requests if they modify them.

mwestacott: Yep. Will amend.

PR #2, opened 2020-06-09
BI-3864 Basic Java Spring Boot application

mwestacott:
Includes a controller implementation of the Swagger spec API.

This is just the Java code. Envisage that other supporting files to faciliate running under UBIC, build with Concourse (e.g. start script, make file) will be added in separate PRs.


pom.xml
```diff
@@ -0,0 +1,120 @@
...
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+        </dependency>
+        
+        <!-- TODO Include this configuration when data is persisted and queried -->
+
+<!--         <dependency> -->
```
mwestacott, line 39: Commented these ones out because they mandated adding DB connection properties, which we don't yet know (or need).

.gitignore
```diff
@@ -21,3 +21,20 @@
...
+.settings
+bin/
+
+# IntelliJ files
+.idea/
+*.iml
+
+# Mac files
+.DS_Store
+._*
```
markpit, line 40: needs a new line at end of file judging by the icon that has appeared

pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    <version>unversioned</version>
+    <packaging>jar</packaging>
+
+    <name>strike-off-objections-api</name>
+    <description>Create a new strike-off objection request</description>
+
+    <parent>
+        <groupId>uk.gov.companieshouse</groupId>
+        <artifactId>companies-house-spring-boot-parent</artifactId>
+        <version>1.3.0</version>
```
markpit, line 16: check this is the latest parent and latest ch dependencies. i know for PTF we weren't on the latest structured logging version

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/PromiseToFileResponseFactory.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+import uk.gov.companieshouse.service.ServiceResult;
+import uk.gov.companieshouse.service.ServiceResultStatus;
+import uk.gov.companieshouse.service.rest.response.ChResponseBody;
+import uk.gov.companieshouse.service.rest.response.ResponseEntityFactory;
+
+
+public class PromiseToFileResponseFactory implements ResponseEntityFactory {
```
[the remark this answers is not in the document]
mwestacott: Oh yes!

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/ApiLogger.java
```diff
@@ -0,0 +1,37 @@
+package uk.gov.companieshouse.api.strikeoffobjections.common;
```
markpit, line 1: would uk.gov.companieshouse.api.strike_off_objections.common be better for the package name
https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html

mwestacott: Agreed. Follows standard so will change.

markpit: The CH standard seems to be what you have so leave as is

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionRequestController.java
```diff
@@ -0,0 +1,77 @@
...
+    public ObjectionRequestController(
+            PluggableResponseEntityFactory responseEntityFactory,
+            ApiLogger apiLogger) {
+
+        this.responseEntityFactory = responseEntityFactory;
+        this.apiLogger = apiLogger;
+    }
+
+    @PostMapping("/request")
+    public ResponseEntity<ChResponseBody<ObjectionResponseBody>> updateCurrentCase(
```
markpit, line 43: how about naming this method createObjectionRequest?

src/main/resources/application.properties
```diff
@@ -0,0 +1,3 @@
+api.endpoint=/company/{companyNumber}/objections/strike-off
```
markpit, line 1: i know we've done it this way for the other apps but i found it a bit of a pain when looking back at them recently, to find out what the base endpoint was. So, would it be better to place this endpoint back into the controller classes to make them more readable - the reader then doesn't have to go and look up this constant value.

mwestacott: Makes sense, will change. Assume there isn't a valid, long lost reason why we did it this way previously.

markpit: I think it was my idea originally in extensions, just thought it was a way of having the base_url in one place , avoids typos etc..  Maybe it is still a good idea, downside is the controllers aren't as readable. Maybe see what the others think?

markpit: Leave it as is, seems a common way of doing it on stack overflow

pom.xml
mwestacott: Yes, this is the latest parent but some of the other dependencies were older. I've updated them and app still builds/starts up/works.

Have noticed that the logging output now looks a bit strange (just more verbose). Might be possible to tweak this, so will investigate but plan to change (if possible) in a separate PR.

PR #3, opened 2020-06-10
Added Makefile and version file

ICoxCH:


Makefile
```diff
@@ -0,0 +1,52 @@
...
+.PHONY: dist
+dist: clean build package
+
+.PHONY: sonar
+sonar:
+	mvn sonar:sonar
+
+.PHONY: sonar-pr-analysis
+sonar-pr-analysis:
+	mvn sonar:sonar -P sonar-pr-analysis
```
markpit, line 52: needs new line on end

version
```diff
@@ -0,0 +1 @@
+0.1
```
markpit, line 1: needs new line on end

Makefile
ICoxCH: Is it a ch standard to have new lines at the end of files?

markpit: github has flagged it with an icon and i think it can cause problems sometimes in shell scripts

PR #6, opened 2020-06-17
BI-3864 Add missing YAML routes file for correct deployment and launch

mwestacott:


routes.yaml
```diff
@@ -0,0 +1,6 @@
+app_name: strike-off-objections-api
+group: internalapi
+
+weight: 900
+routes:
+  1: ^/company/(.*)/strike-off-objections/*
```
mattch1, line 6: newline

mwestacott: Gets me every time!

mwestacott: Done.

PR #7, opened 2020-06-29
BI-4109 create swagger spec

markpit:
Ditched the idea for multiple reasons. As we have to use HATEOS links, we would need new reason endpoints to get and add reasons.
Not sure if Patch should return the resource - at the moment it is.
Some http status' might not be correct


spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
+        "summary": "Create a new strike-off objection request",
         "produces": [
           "application/json"
         ],
+        "tags": [
+          "Strike Off Objections"
+        ],
+        "responses": {
+          "201": {
+            "description": "An objection request has been received.",
```
mwestacott, line 52: Objection request has been 'created' (as well as 'received').

spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
+            "schema": {
+              "$ref": "#/definitions/ResourceCreated"
```
mwestacott, line 54: Could be more specific here - 'ObjectionCreated' maybe.

markpit: the same object structure is returned by post attachment as well so I named it ResourceCreated for re-use. I could name it ObjectionCreated but I'd also have to create an AttachmentCreated as well.

spec/schema.json
markpit: ok

spec/schema.json
mwestacott: Ok. Fair enough. Confluence page had different field names being returned (objectionId and fileId) but I'm happy with the generic approach.

spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
+        {
+          "name": "objectionId",
+          "in": "path",
+          "required": true,
+          "description": "ID of the Objection",
+          "type": "string"
+        }
+      ],
+      "get": {
+        "summary": "Returns the objection request identified by objectionId",
```
mwestacott, line 71: Not an 'objection request' here, just an 'objection'.

spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
+          "404": {
+            "description": "not found"
+          }
+        }
+      }
+    },
+    "/": {
       "post": {
-        "summary": "Temporary end-point to prove sprint zero connectivity from the web app",
+        "summary": "Create a new strike-off objection request",
```
mwestacott, line 43: I think 'request' should go here. In a few other places too. HTTP requests are coming in, as you pointed out, but our terminology now doesn't have 'objection requests'.

spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
+        "summary": "Returns the objection request identified by objectionId",
+        "produces": [
+          "application/json"
+        ],
+        "tags": [
+          "Strike Off Objections"
+        ],
+        "responses": {
+          "200": {
+            "description": "a Status object has been returned successfully",
```
mwestacott, line 80: 'An objection has been returned successfully'.

spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
             "schema": {
-              "$ref": "#/definitions/ObjectionOutput"
+              "$ref": "#/definitions/ResourceCreated"
             }
+          },
+          "400": {
+            "description": "there was an error processing the request"
+          },
+          "404": {
+            "description": "the request or reason does not exist"
```
mwestacott, line 140: I think this should be 'the objection does not exist'.

spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
+        }
+      }
+    },
+    "CreatedBy": {
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "string"
+        },
+        "forename": {
```
mwestacott, line 330: Where / how will we get forename and surname from? I thought that user just signs in with eMail address.

spec/schema.json
markpit: updated

spec/schema.json
markpit: updated

spec/schema.json
```diff
@@ -16,53 +16,346 @@
...
+        },
+        "status": {
+          "$ref": "#/definitions/Status"
         }
       }
+    },
+    "Status": {
+      "type": "string",
+      "enum": [
+        "open",
```
mwestacott, line 356: Personally I like enum values in uppercase. I just think it helps them stand out (a bit like a constant) but up to you.

spec/schema.json
markpit: updated

spec/schema.json
markpit: updated

spec/schema.json
markpit: we get them from ERIC, I guess we don't technically need them as we can cross reference the user data

mwestacott: Fair enough if we have them.

markpit: i've taken them out, they were just nice to have's

spec/schema.json
markpit: updated

PR #9, opened 2020-06-30
BI-4109 api spec update, adding GET for all attachments ad updating h…

markpit:
…ostname

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -5,7 +5,7 @@
     "version": "1.0.0",
     "title": "Companies House - Strike Off Objections API"
   },
-  "host": "internalapi.companieshouse.gov.uk",
+  "host": "api.companieshouse.gov.uk",
   "basePath": "/company/{companyNumber}/strike-off-objections",
   "schemes": [
     "https"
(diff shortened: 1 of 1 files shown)


spec/schema.json
```diff
@@ -118,6 +118,29 @@
...
+            "description": "Attachments returned successfully",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/Attachment"
+              }
+            }
+          },
+          "404": {
+            "description": "the objection or attachments were not found"
```
ICoxCH, line 140: This to me suggests that if an objection has no attatchments, it will return a 404.
Would returning a 200 with an empty array be clearer?

markpit: it does look better, i'll update

spec/schema.json
```diff
@@ -118,6 +118,29 @@
...
+            "description": "Attachments returned successfully",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/Attachment"
+              }
+            }
+          },
+          "404": {
+            "description": "the objection was not found"
```
markpit, line 140: I'll capitalise the O

PR #10, opened 2020-06-30
BI-4125 Rearranged field to match schema

mattch1:
As we have agreed on a single field for the reason there is not a lot to do - having looked at the schema have rearraged order of fields to match it.

One other change that could be made is to change the entity name form "StrikeOffObjectionsEntity" to "Objection".

Also do we nned to change the annoation @Document(collection = "strike_off_objections")?


markpit:
personally, entity name would be better as Objection
document name looks to be in line with what Martin has created so no change needed

mattch1:
This is what's in the spec so renamed to "Objection".

PR #11, opened 2020-06-30
bi 4116 create objection

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionRequestController.java
```diff
@@ -1,77 +1,75 @@
...
-            // TODO Implement actual business logic here when stories are available
-            
-            ObjectionResponseBody objectionResponseBody = new ObjectionResponseBody();
-            objectionResponseBody.setRequestId("12345");
-            
-            return responseEntityFactory.createResponse(ServiceResult.created(objectionResponseBody));
-
-        } catch (Exception e) {            
+            String objectionId = objectionService.createObjection(requestId, companyNumber);
+            ObjectionResponse response = new ObjectionResponse(objectionId);
```
markpit, line 56: do we need to create an ObjectionResponse object or can we just return the string?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public String createObjection(String requestId, String companyNumber) {
+        Map<String, Object> logMap = new HashMap<>();
+        logMap.put(LogConstants.COMPANY_NUMBER.getValue(), companyNumber);
+        logger.infoContext(requestId, "Creating objection", logMap);
+
+        StrikeOffObjectionsEntity entity = new StrikeOffObjectionsEntity.Builder()
+                .withCompanyNumber(companyNumber)
+                .withCreatedOn(LocalDateTime.now())
```
markpit, line 35: this is where the date supplier would be handy, we can then mock the supplier to provide a date and we can check that the entity is populated with the correct timestamp.

src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+spring.data.mongodb.uri=${MONGODB_URL}/strike_off_objections
```
markpit, line 1: github is moaning about newlines at end of a few of your files.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionRequestController.java
ICoxCH: Just returning the string wouldn't conform to the spec which wants `"id" : "OBJECTION_ID"`
I could possibly return a map to achieve the same effect

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+    ObjectionService objectionService;
+
+    @Test
+    void createObjectionTest() {
+        StrikeOffObjectionsEntity returnedEntity = new StrikeOffObjectionsEntity.Builder()
+                .withCompanyNumber(COMPANY_NUMBER)
+                .build();
+        returnedEntity.setId(OBJECTION_ID);
+        when(strikeOffObjectionsRepository.save(any())).thenReturn(returnedEntity);
+
```
markpit, line 39: could you mock the builder and check it is called with the correct company number and current timestamp (supplier?)
Then you could confirm that the entity returned from the builder is the same one passed into the save() call

ICoxCH: I thought as the builder is static, Mockito cannot mock it?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionRequestController.java
markpit: ok I forgot about the 2 parts

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
markpit: ok, but we need a way to check the entity is created with correct values. If we can't mock the builder, let it build a real one, capture the argument passed in to save() and check it has the same company number that was passed in to the method and that the timestamp is correct. 
You could still have a dummy response returned from save() and check the id is being returned from createObjection (as it's currently doing)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/StrikeOffObjectionsEntity.java
```diff
@@ -16,6 +16,7 @@
         private CreatedBy createdBy;
         private LocalDateTime createdOn;
         private ObjectionStatus status;
+        private String httpRequestId;
```
markpit, line 19: can you annotate httpRequestId with @org.springframework.data.annotation.Transient
i haven't tried it, but it should mean that it will be ignored when we convert this to Json to send back to client.
https://stackoverflow.com/a/42753469

ICoxCH: Will try that now

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
```diff
@@ -19,26 +24,34 @@
...
-        when(strikeOffObjectionsRepository.save(any())).thenReturn(returnedEntity);
+        when(objectionRepository.save(any())).thenReturn(returnedEntity);
+        when(localDateTimeSupplier.get()).thenReturn(MOCKED_TIME_STAMP);
 
+        ArgumentCaptor<Objection> acObjection = ArgumentCaptor.forClass(Objection.class);
         String returnedId = objectionService.createObjection(REQUEST_ID, COMPANY_NUMBER);
 
+        verify(objectionRepository).save(acObjection.capture());
         assertEquals(OBJECTION_ID, returnedId);
+        assertEquals(MOCKED_TIME_STAMP, acObjection.getValue().getCreatedOn());
```
markpit, line 55: also assert acObjection has had the COMPANY_NUMBER set on it (to ensure the number passed into createObjection is being set on the entity)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionRequestController.java
```diff
@@ -1,77 +1,75 @@
...
+            return responseEntityFactory.createResponse(ServiceResult.created(response));
+        } catch (Exception e) {
             apiLogger.errorContext(
-                requestId,
-                String.format("Error processing the Strike-Off Objection request for %s %s",
-                    LOG_COMPANY_NUMBER_KEY, companyNumber),
-                e);
+                    requestId,
+                    "Error processing the Strike-Off Objection request",
```
mwestacott, line 61: "Error creating the strike-off objection request" maybe? Just a bit more specific. ('request' at end could be dropped too)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionRequestController.java
```diff
@@ -1,77 +1,75 @@
...
 
-            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+            return responseEntityFactory.createEmptyInternalServerError();
         } finally {
             apiLogger.infoContext(
-                requestId,
-                String.format("Finished POST /current request: %s %s, %s %s",
-                    LOG_REQUEST_ID_KEY, requestId,
-                    LOG_COMPANY_NUMBER_KEY, companyNumber));
+                    requestId,
```
mwestacott, line 69: Nice formatting 👍 .

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+        ArgumentCaptor<Objection> acObjection = ArgumentCaptor.forClass(Objection.class);
+        String returnedId = objectionService.createObjection(REQUEST_ID, COMPANY_NUMBER);
+
+        verify(objectionRepository).save(acObjection.capture());
+        assertEquals(OBJECTION_ID, returnedId);
+        assertEquals(MOCKED_TIME_STAMP, acObjection.getValue().getCreatedOn());
+        assertEquals(COMPANY_NUMBER, acObjection.getValue().getCompanyNumber());
+    }
+}
```
mattch1, line 58: new line

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionRequestControllerTest.java
```diff
@@ -0,0 +1,71 @@
...
+        when(pluggableResponseEntityFactory.createEmptyInternalServerError()).thenReturn(
+                ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build()
+        );
+
+        ResponseEntity<ChResponseBody<ObjectionResponse>> response = objectionRequestController.createObjection(COMPANY_NUMBER, REQUEST_ID);
+
+        assertEquals(HttpStatus.INTERNAL_SERVER_ERROR, response.getStatusCode());
+
+    }
+}
```
mattch1, line 71: new line

PR #12, opened 2020-06-30
BI-4109 Api spec - adding http_request_id to help support team trace …

markpit:
…in logs

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -267,6 +267,9 @@
           "type": "string",
           "format": "uuid"
         },
+        "http_request_id": {
+          "type": "string"
+        },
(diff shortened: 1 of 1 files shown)


spec/schema.json
```diff
@@ -267,6 +267,9 @@
           "type": "string",
           "format": "uuid"
         },
+        "http_request_id": {
```
mwestacott, line 270: Is this needed / ever going to be used by the client? Understand that it will be stored in Mongo DB but do we need to expose this internal information through the public interface?

markpit: i guess not. i'll close.

PR #18, opened 2020-07-07
obj-63-add-attachment-to-api

mattch1:
Borrowed heavily from extensions-api - have refactored out the annonymous calls in the client into their own methods so it is more readable and clearer what the objects returned are. Moreover, updated unit tests to use junit jupiter. Also added headers to the response object in line with the acceptance criteria.

So far only been able to test with unit tests.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/ApiLogger.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/ApiLogger.java
@@ -31,10 +31,6 @@ public class ApiLogger {
         LOGGER.errorContext(context, e, null);
     }
 
-    public void errorContext(String context) {
-        LOGGER.errorContext(context, null, null);
-    }
-
     public void errorContext(String context, String message, Exception e) {
         LOGGER.errorContext(context, message, e, null);
     }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/ApiLogger.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/ApiLogger.java
@@ -31,6 +31,10 @@ public class ApiLogger {
         LOGGER.errorContext(context, e, null);
     }
 
+    public void errorContext(String context) {
+        LOGGER.errorContext(context, null, null);
+    }
+
     public void errorContext(String context, String message, Exception e) {
         LOGGER.errorContext(context, message, e, null);
     }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -121,7 +121,7 @@ public class ObjectionController {
 
             apiLogger.errorContext(
                     requestId,
-                    "Objection not found",
+                    OBJECTION_NOT_FOUND,
                     e,
                     logMap
             );
@@ -163,7 +163,7 @@ public class ObjectionController {
 
             apiLogger.errorContext(
                     requestId,
-                    "Objection not found",
+                    OBJECTION_NOT_FOUND,
                     e,
                     logMap
             );
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -195,7 +195,7 @@ public class ObjectionController {
         );
 
         try {
-            ServiceResult<String> result = objectionService.addAttachment(objectionId, file);
+            ServiceResult<String> result = objectionService.addAttachment(requestId, file);
             return new ResponseEntity(result.getData(), HttpStatus.OK);
         } catch(ServiceException e) {
 
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -131,14 +131,15 @@ public class ObjectionController {
     public ResponseEntity<String> uploadAttachmentToObjection (
             @RequestParam("file") MultipartFile file,
             @PathVariable("companyNumber") String companyNumber,
-            @PathVariable String objectionId) {
+            @PathVariable String objectionId,
+            @RequestHeader(value = ERIC_REQUEST_ID_HEADER) String requestId) {
 
         Map<String, Object> logMap = new HashMap<>();
         logMap.put(LOG_COMPANY_NUMBER_KEY, companyNumber);
         logMap.put(LOG_OBJECTION_ID_KEY, objectionId);
 
         apiLogger.infoContext(
-                objectionId,
+                requestId,
                 "POST /{objectionId}/attachments request received",
                 logMap
         );
@@ -149,7 +150,7 @@ public class ObjectionController {
         } catch(ServiceException e) {
 
             apiLogger.errorContext(
-                    objectionId,
+                    requestId,
                     "Objection not found",
                     e,
                     logMap
@@ -157,7 +158,7 @@ public class ObjectionController {
             return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
         } finally {
             apiLogger.infoContext(
-                    objectionId,
+                     requestId,
                     "Finished POST /{objectionId}/attachments request",
                     logMap
             );
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -18,8 +18,6 @@ import org.springframework.web.multipart.MultipartFile;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
 import uk.gov.companieshouse.api.strikeoffobjections.common.LogConstants;
 import uk.gov.companieshouse.api.strikeoffobjections.exception.ObjectionNotFoundException;
-import uk.gov.companieshouse.api.strikeoffobjections.file.FileTransferApiClient;
-import uk.gov.companieshouse.api.strikeoffobjections.file.FileTransferApiClientResponse;
 import uk.gov.companieshouse.api.strikeoffobjections.model.patch.ObjectionPatch;
 import uk.gov.companieshouse.api.strikeoffobjections.model.response.ObjectionResponse;
 import uk.gov.companieshouse.api.strikeoffobjections.service.IObjectionService;
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -9,14 +9,19 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
+import org.springframework.web.multipart.MultipartFile;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
 import uk.gov.companieshouse.api.strikeoffobjections.common.LogConstants;
 import uk.gov.companieshouse.api.strikeoffobjections.exception.ObjectionNotFoundException;
+import uk.gov.companieshouse.api.strikeoffobjections.file.FileTransferApiClient;
+import uk.gov.companieshouse.api.strikeoffobjections.file.FileTransferApiClientResponse;
 import uk.gov.companieshouse.api.strikeoffobjections.model.patch.ObjectionPatch;
 import uk.gov.companieshouse.api.strikeoffobjections.model.response.ObjectionResponse;
 import uk.gov.companieshouse.api.strikeoffobjections.service.IObjectionService;
+import uk.gov.companieshouse.service.ServiceException;
 import uk.gov.companieshouse.service.ServiceResult;
 import uk.gov.companieshouse.service.rest.response.ChResponseBody;
 import uk.gov.companieshouse.service.rest.response.PluggableResponseEntityFactory;
@@ -34,10 +39,13 @@ public class ObjectionController {
 
     private PluggableResponseEntityFactory responseEntityFactory;
     private IObjectionService objectionService;
+
     private ApiLogger apiLogger;
 
     @Autowired
-    public ObjectionController(PluggableResponseEntityFactory responseEntityFactory, IObjectionService objectionService, ApiLogger apiLogger) {
+    public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
+                               IObjectionService objectionService,
+                               ApiLogger apiLogger) {
         this.responseEntityFactory = responseEntityFactory;
         this.objectionService = objectionService;
         this.apiLogger = apiLogger;
@@ -116,6 +124,18 @@ public class ObjectionController {
                     "Finished PATCH /{objectionId} request",
                     logMap
             );
+        }
+    }
+
+    @PostMapping("/{objectionId}/attachments")
+    public ResponseEntity<String> uploadAttachmentToRequest(
+            @RequestParam("file") MultipartFile file, @PathVariable String objectionId) {
+        try {
+            ServiceResult<String> result = objectionService.addAttachment(objectionId, file);
+            return new ResponseEntity(result, HttpStatus.OK);
+        } catch(ServiceException e) {
+
+            return ResponseEntity.status(503).build();
         }
     }
 }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -128,14 +128,39 @@ public class ObjectionController {
     }
 
     @PostMapping("/{objectionId}/attachments")
-    public ResponseEntity<String> uploadAttachmentToRequest(
-            @RequestParam("file") MultipartFile file, @PathVariable String objectionId) {
+    public ResponseEntity<String> uploadAttachmentToObjection (
+            @RequestParam("file") MultipartFile file,
+            @PathVariable("companyNumber") String companyNumber,
+            @PathVariable String objectionId) {
+
+        Map<String, Object> logMap = new HashMap<>();
+        logMap.put(LOG_COMPANY_NUMBER_KEY, companyNumber);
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/ApiLogger.java
```diff
@@ -31,6 +31,10 @@ public void errorContext(String context, Exception e) {
         LOGGER.errorContext(context, e, null);
     }
 
+    public void errorContext(String context) {
+        LOGGER.errorContext(context, null, null);
```
ICoxCH, line 35: I'm not sure what the purpose of this method would be?
Context is the requestId so it would just be an error message with a request id with no info on what went on?
I would reserve error calls for when there is an exception, perhaps define and throw a custom exception and add that to the log call

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private <T> FileTransferApiClientResponse makeApiCall(FileTransferOperation <T> operation,
+                                                          FileTransferResponseBuilder <T> responseBuilder) {
+        FileTransferApiClientResponse response = new FileTransferApiClientResponse();
+
+        try {
+            T operationResponse = operation.execute();
+            response = responseBuilder.createResponse(operationResponse);
+        } catch (IOException e) {
+            logger.errorContext(e.getMessage(), e);
```
ICoxCH, line 45: As briefly mentioned above, the context argument is meant to be the request id of the request.
I would suggest passing the request id into this method and adding it as the context argument
logger.errorContext(requestId, e.getMessage(), e);

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,125 @@
...
+        FileTransferApiClientResponse fileTransferApiClientResponse = new FileTransferApiClientResponse();
+        if (responseEntity != null) {
+            fileTransferApiClientResponse.setHeaders(responseEntity.getHeaders());
+            fileTransferApiClientResponse.setHttpStatus(responseEntity.getStatusCode());
+            FileTransferApiResponse apiResponse = responseEntity.getBody();
+            if (apiResponse != null) {
+                fileTransferApiClientResponse.setFileId(apiResponse.getId());
+            }
+        } else {
+            logger.errorContext(String.format("%s %s", NULL_RESPONSE_MESSAGE,fileTransferApiURL));
```
ICoxCH, line 96: Again, I would ensure the requestId is passed in as the context argument.
As this is 'expected' behavior, I don't thi k we need an errorContext here, infoContext(context, message) would be fine

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public FileTransferApiClientResponse upload(MultipartFile fileToUpload) {
+        return makeApiCall(
+             () -> getFileTransferOperation(fileToUpload),
+             responseEntity ->  getFileTransferApiClientResponse(responseEntity)
+        );
+    }
+
+    /**
+     * @param fileToUpload
```
mwestacott, line 67: Javadoc is incomplete.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,125 @@
...
+        LinkedMultiValueMap<String, Object> body = createUploadBody(fileHttpEntity);
+        HttpEntity<LinkedMultiValueMap<String, Object>> requestEntity = new HttpEntity<>(body, headers);
+        return restTemplate.postForEntity(fileTransferApiURL, requestEntity, FileTransferApiResponse.class);
+    }
+
+    /**
+     * FileTransferResponseBuilder - the output from FileTransferOperation is the input into
+     * this FileTransferResponseBuilder
+     * @param responseEntity
+     * @return
```
mwestacott, line 84: Return object should be described.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClientResponse.java
```diff
@@ -0,0 +1,35 @@
+package uk.gov.companieshouse.api.strikeoffobjections.file;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+
+public class FileTransferApiClientResponse {
+
+    private String fileId;
+    private HttpStatus httpStatus;
+    private HttpHeaders headers;
```
mwestacott, line 10: httpHeaders for this member variable name, for consistency with httpStatus.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/ApiLogger.java
mattch1: Removed this method

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Added a request id

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Completed this

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Description added

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClientResponse.java
mattch1: Renamed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Changed to info context.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Uploads a file to the file-transfer-api
+     * Creates a multipart form request containing the file and sends to
+     * the file-transfer-api. The response from the file-transfer-api contains
+     * the new unique id for the file. This is captured and returned in the FileTransferApiClientResponse.
+     * @param fileToUpload The file to upload
+     * @return FileTransferApiClientResponse containing the file id if successful, the http header and http status
+     */
+    public FileTransferApiClientResponse upload(String requestId, MultipartFile fileToUpload) {
+        this.requestId = requestId;
```
ICoxCH, line 62: Can we guarentee that only one request will be being executed by this object at once?
As if multiple requests go through the object at the same time, then all their request id's will overwrite and interfere with each other.

If each new request instantiates a new FileTransferApiClient that will be ok.

Though would it be better if FileTransferApiClient was a Component that was instantiated by spring and was autowired into services that need it? In which case I would suggest passing the requestId in as method arguments where needed to avoid having "state"

ICoxCH:
More of a general comment here but the branch name, and story it is linked to suggest that an endpoint should be being created but it doesn't seem there is one in the pr

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Fair point - it is risky  so removed this.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,131 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.multipart.MultipartFile;
+import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
+
+import java.io.IOException;
+
+public class FileTransferApiClient {
```
markpit, line 17: I think ian mentioned this already but this can be a component as in extensions

mattch1:
This is more an implementation of an endpoint and soemthing that contacts a further endpoint rather than being an enpoint in itself.

markpit:
> Borrowed heavily from extensions-api - have refactored out the annonymous calls in the client into their own methods so it is more readable and clearer what the objects returned are. Moreover, updated unit tests to use junit jupiter. Also added headers to the response object in line with the acceptance criteria.
> 
> So far only been able to test with unit tests.

You should make sure this runs before merging, as our pipelines will deploy straight into staging soon and then live eventually. You should be able to test this by using postman to hit the endpoint.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Yep, doing that next.

mattch1:
We should test this through postman - do we have an attachments endpoint yet?

markpit:
> > Borrowed heavily from extensions-api - have refactored out the annonymous calls in the client into their own methods so it is more readable and clearer what the objects returned are. Moreover, updated unit tests to use junit jupiter. Also added headers to the response object in line with the acceptance criteria.
> > So far only been able to test with unit tests.
> 
> You should make sure this runs before merging, as our pipelines will deploy straight into staging soon and then live eventually. You should be able to test this by using postman to hit the endpoint.
If this isn't tied to the endpoint yet then I guess you can't run it end to end. In which case unit testing should be enough. But we'd need to test it when it gets integrated with the endpoint

markpit:
> We should test this through postman - do we have an attachments endpoint yet?

I thought this was to implement the endpoint, if they are not connected yet then you won't be able to test it via postman yet.

ICoxCH:
> We should test this through postman - do we have an attachments endpoint yet?

From the description of this story:

Send incoming file to the file-transfer-api

return the file id returned from file-transfer-api

Return error codes returned from file-transfer-api? do we return the invalid file type http code at this point?

return header information in response header containing file metadata


Im reading as this story you should be able to call the endpoint, and receive the file id in the response

markpit:
> > We should test this through postman - do we have an attachments endpoint yet?
> 
> From the description of this story:
> 
> Send incoming file to the file-transfer-api
> 
> return the file id returned from file-transfer-api
> 
> Return error codes returned from file-transfer-api? do we return the invalid file type http code at this point?
> 
> return header information in response header containing file metadata
> 
> Im reading as this story you should be able to call the endpoint, and receive the file id in the response

Going forward, could we have multiple PR's for one story? It might be better in some cases to commit isolated components and then link them up later?

mattch1:
I did wonder about this - but  I wanted to avoid doing too much in one PR because of problems  have had in the past.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,133 @@
...
+import org.springframework.web.multipart.MultipartFile;
+import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
+
+import java.io.IOException;
+
+@Component
+public class FileTransferApiClient {
+
+    private static final String HEADER_API_KEY = "x-api-key";
+    private static final String UPLOAD = "upload";
```
markpit, line 22: It was probably already like this in extensions, but I think UPLOAD_FILE_FIELD_NAME = "upload" would be more meaningful

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -118,4 +126,16 @@ public ResponseEntity patchObjection(
             );
         }
     }
+
+    @PostMapping("/{objectionId}/attachments")
+    public ResponseEntity<String> uploadAttachmentToRequest(
```
markpit, line 131: For the method name, we seem to be referring to the requests as Objection in other methods i noticed. So I would either just call it uploadAttachment or uploadAttachmentToObjection

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -118,4 +126,16 @@ public ResponseEntity patchObjection(
...
+            @RequestParam("file") MultipartFile file, @PathVariable String objectionId) {
+        try {
+            ServiceResult<String> result = objectionService.addAttachment(objectionId, file);
```
ICoxCH, line 134: You seem to be using the objectionId as the requestId.
The objectionId is the id of the objection object in the mongo database
The request Id is generated for each request by eric. We want to be using the requestId from eric as the context argument in all the log statements.
Though It is definitely worth adding the objectionId to the dataMap for the log statement for more information.

If you look at the patchObjection method it shows how to use the data map in the log, and how to get the requestId easily

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -67,4 +80,20 @@ public void patchObjection(String requestId, String companyNumber, String object
...
+    public ServiceResult<String> addAttachment(String objectionId, MultipartFile file) throws ServiceException {
+        FileTransferApiClientResponse response = fileTransferApiClient.upload(objectionId, file);
+
+        HttpStatus responseHttpStatus = response.getHttpStatus();
+        if (responseHttpStatus != null && responseHttpStatus.isError()) {
+            throw new ServiceException(responseHttpStatus.toString());
+        }
+        String fileId = response.getFileId();
+        if (StringUtils.isBlank(fileId)) {
+            throw new ServiceException("No file id returned from file upload");
```
markpit, line 93: I noticed in extensions-api attachments controller.upload endpoint, it is catching
catch(HttpClientErrorException | HttpServerErrorException e) 

assuming the code is similar, we would need to handle these exceptions either here or somewhere further up the chain possibly in the fileTransferClient

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: Using data map logging: 

2020-07-08T14:42:28.591Z [8waDSJX3Q2g-2NZM2qQ211SeYzZ5] info: POST /{objectionId}/attachments request received

ZM2qQ211SeYzZ5] info: Finished POST /{objectionId}/attachments request

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: Method name changed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Changed name.

markpit:
looks ok but sonar is failing for unit test coverage

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mwestacott: @return still has no information.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -0,0 +1,133 @@
...
+             responseEntity ->  getFileTransferApiClientResponse(requestId, responseEntity)
+        );
+    }
+
+    /**
+     * Calls the file transfer api and returns the result of the
+     * call ready to pass to the response builder
+     * @param fileToUpload
+     * @return ResponseEntity containing the raw data from which the response object is built
+     * @throws IOException
```
mwestacott, line 76: Since you've included the @throws Javadoc tag you should add some words around when that exception might be thrown.

mwestacott:
Please just have a look at my Javadoc comments - an earlier one hasn't been addressed and I've added a new one. Code looks fine though.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Added information.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
mattch1: Added information to return

mattch1:
Updated javadocs added some tests can only add more once PR https://github.com/companieshouse/strike-off-objections-api/pull/20/files is merged

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -67,4 +80,20 @@ public void patchObjection(String requestId, String companyNumber, String object
         }
     }
 
+    @Override
+    public ServiceResult<String> addAttachment(String objectionId, MultipartFile file) throws ServiceException {
```
ICoxCH, line 84: Sorry if it seems I keep banging on about it, you have passed the objectionId from the controller into this method here, but are using the objetionId as the requestId argument in fileTransferApiClient.upload() on line 85.
I think you should be passing the requestId in as this param and then passing the requestId into the fileTransferApiClient.upload()

markpit: maybe rename requestId to httpRequestId to clear it up as in Extensions the Request was the entity, but now Objection is the entity.

mattch1: Renaming the requestId a good idea but I'd rather put it to one side for now as it relates to several areas of the code. I'll stick with requestId for now.

mattch1: Changed this.

mattch1:
Mistakenly thought ide had done this for all insatnces - changed them manually.

PR #24, opened 2020-07-10
OBJ-65 Ensure correct JSON response content is returned (to match spec)

mwestacott:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/AttachmentResponseDTO.java
```diff
@@ -1,19 +1,36 @@
...
-    
+
+    @JsonProperty("content_type")
     private String contentType;
     
+    @JsonProperty("size")
     private long size;
 
+    @JsonProperty("links")
+    @JsonUnwrapped
```
markpit, line 23: with json unwrapped i thought it inlined that object.
i think the json should look like 
{
  name,
  etc..
  links {
    link1,
    etc..
  }

mwestacott: Without the JsonUnwrapped we were getting 'links' twice:

"links":
	{"links":{"download":"/company/00006400/strike-off-objections/5f05c3f24be29647ef076f21/attachments/123-456-789/download",
          "self":"/company/00006400/strike-off-objections/5f05c3f24be29647ef076f21/attachments/123-456-789"}
}

Adding it fixes it.

So output now looks like:

[{"id":"123-456-789",
"name":"original-file-name.txt",
"content_type":"TEXT-FILE",
"size":44,
"links":{"download":"/company/00006400/strike-off-objections/5f05c3f24be29647ef076f21/attachments/123-456-789/download","self":"/company/00006400/strike-off-objections/5f05c3f24be29647ef076f21/attachments/123-456-789"}}]

markpit: ok that's fine. I think i may have told matt to take it off before. didn't realise it was a 'twice' issue.

PR #27, opened 2020-07-14
Feature/obj 118 attachments wrong status code

mattch1:
Changed method to return 200 rather than 201.


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -163,7 +163,7 @@ public ResponseEntity patchObjection(
             List<AttachmentResponseDTO> attachmentResponseDTOs = attachments.stream()
                     .map(attachmentMapper::attachmentEntityToAttachmentResponseDTO).collect(Collectors.toList());
 
-            return responseEntityFactory.createResponse(ServiceResult.created(attachmentResponseDTOs));
+            return responseEntityFactory.createResponse(ServiceResult.found(attachmentResponseDTOs));
```
ICoxCH, line 166: does found return a 200? Wouldn't we want it to be OK?

mattch1: It return 200 checked with Tom Strickland and tested this.
Screen Shot 2020-07-14 at 12 29 27

PR #30, opened 2020-07-20
OBJ-134 File Transfer Api call for delete

mattch1:
Call to the file transfer api to removed the file from s3 will proceeed with deletion from mongo if s3 deletion is sucessful so as to avoid leaving the objection in an inconsistent state.

Added unit tests for the service and file transfer client.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -49,6 +49,7 @@ public class ObjectionController {
     private static final String ERIC_AUTHORISED_USER = "ERIC-Authorised-User";
     private static final String OBJECTION_NOT_FOUND = "Objection not found";
     private static final String ATTACHMENT_NOT_FOUND = "Attachment not found";
+    public static final String COULD_NOT_DELETE = "Could not delete attachment";
 
     private PluggableResponseEntityFactory responseEntityFactory;
     private IObjectionService objectionService;
@@ -336,6 +337,17 @@ public class ObjectionController {
             );
 
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+
+        } catch(ServiceException e) {
+
+            apiLogger.errorContext(
+                    requestId,
+                    COULD_NOT_DELETE,
+                    e,
+                    logMap
+            );
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+
         } finally {
             apiLogger.infoContext(
                     requestId,
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -316,7 +316,7 @@ public class ObjectionController {
         );
 
         try {
-            objectionService.deleteAttachment(requestId, companyNumber, objectionId, attachmentId);
+            objectionService.deleteAttachment(requestId, objectionId, attachmentId);
 
             return ResponseEntity.noContent().build();
         } catch (ObjectionNotFoundException e) {
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
@@ -40,8 +40,8 @@ public class ObjectionService implements IObjectionService {
 
     private static final String OBJECTION_NOT_FOUND_MESSAGE = "Objection with id: %s, not found";
     private static final String ATTACHMENT_NOT_FOUND_MESSAGE = "Attachment with id: %s, not found";
-    private static final String DELETE_ERROR_MESSAGE = "Unable to delete attachment %s, status code %s";
-    private static final String DELETE_ERROR_MESSAGE_SHORT = "Unable to delete attachment %s";
+    private static final String DELETE_ATTACHMENT_MESSAGE = "Unable to delete attachment %s, status code %s";
+    private static final String DELETE_ATTACHMENT_MESSAGE_SHORT = "Unable to delete attachment %s";
 
     private ObjectionRepository objectionRepository;
     private ApiLogger logger;
@@ -209,19 +209,19 @@ public class ObjectionService implements IObjectionService {
         try {
             FileTransferApiClientResponse response = fileTransferApiClient.delete(requestId, attachmentId);
             if (response == null || response.getHttpStatus() == null) {
-                String message = String.format(DELETE_ERROR_MESSAGE_SHORT, attachmentId);
+                String message = String.format(DELETE_ATTACHMENT_MESSAGE_SHORT, attachmentId);
                 logger.infoContext(requestId, message, logMap);
                 throw new ServiceException(message);
             } else {
                 if (response.getHttpStatus().isError()) {
-                    String message = String.format(DELETE_ERROR_MESSAGE,
+                    String message = String.format(DELETE_ATTACHMENT_MESSAGE,
                             attachmentId, response.getHttpStatus());
                     logger.infoContext(requestId, message, logMap);
                     throw new ServiceException(message);
                 }
             }
         } catch (HttpClientErrorException | HttpServerErrorException e) {
-            String message = String.format(DELETE_ERROR_MESSAGE, attachmentId, e.getStatusCode());
+            String message = String.format(DELETE_ATTACHMENT_MESSAGE, attachmentId, e.getStatusCode());
             logger.errorContext(requestId, message, e, logMap);
             throw new ServiceException(message);
         }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
@@ -5,6 +5,8 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.HttpServerErrorException;
 import org.springframework.web.multipart.MultipartFile;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
 import uk.gov.companieshouse.api.strikeoffobjections.common.LogConstants;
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -196,10 +198,41 @@ public void deleteAttachment(String requestId, String companyNumber, String obje
                 () -> new AttachmentNotFoundException(String.format(ATTACHMENT_NOT_FOUND_MESSAGE, attachmentId))
         );
 
-        attachments.remove(attachment);
+        boolean successfulDelete = deleteFromS3(requestId, companyNumber, objectionId, attachmentId);
 
-        objection.setAttachments(attachments);
+        if(successfulDelete) {
```
ICoxCH, line 203: if delete is unsuccessful we should throw an exception so that the controller can return a 500. As it is an unsuccessful deletion would return a 204 NO CONTENT

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClientUnitTest.java
```diff
@@ -91,6 +93,40 @@ public void testUpload_GenericExceptionResponse() {
         assertThrows(RestClientException.class, () -> fileTransferApiClient.upload(REQUEST_ID, file));
     }
 
+    @Test
+    public void testDelete_success() {
```
ICoxCH, line 97: could these be camelCase to match CH java standards

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -196,10 +198,41 @@ public void deleteAttachment(String requestId, String companyNumber, String obje
...
+
+            objectionRepository.save(objection);
+        }
+    }
+
+    private boolean deleteFromS3(String requestId, String companyNumber, String objectionId, String attachmentId) {
+        Map<String, Object> logMap = new HashMap<>();
+        logMap.put(LogConstants.COMPANY_NUMBER.getValue(), companyNumber);
+        logMap.put(LogConstants.OBJECTION_ID.getValue(), objectionId);
+        final String errorMessage = "Unable to delete attachment %s, status code %s";
```
mwestacott, line 216: Would suggest defining these formatted error messages as class constants rather than creating new strings each time the method is called. A bit like in the FileTransferApiClient class above.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -196,10 +198,41 @@ public void deleteAttachment(String requestId, String companyNumber, String obje
...
+            attachments.remove(attachment);
 
-        objectionRepository.save(objection);
+            objection.setAttachments(attachments);
+
+            objectionRepository.save(objection);
+        }
+    }
+
+    private boolean deleteFromS3(String requestId, String companyNumber, String objectionId, String attachmentId) {
```
mwestacott, line 212: Looks like companyNumber and objectionId aren't needed/used in this method. I think these parameters can just be ommitted.

mattch1: They are needed for the log map

mwestacott: But all we do is log them. They aren't actually used or needed by the method.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -196,10 +195,50 @@ public void deleteAttachment(String requestId, String companyNumber, String obje
...
+            objectionRepository.save(objection);
+        } else {
+            throw new ServiceException(UNABLE_TO_DELETE_ATTACHMENT);
+        }
+    }
 
-        objection.setAttachments(attachments);
+    private boolean deleteFromS3(String requestId, String attachmentId, Map<String, Object>  logMap) {
+        try {
+            FileTransferApiClientResponse response = fileTransferApiClient.delete(requestId, attachmentId);
```
markpit, line 214: rather than return lots of true/false it would be better to throw an exception if unsuccessful

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/utils/Utils.java
```diff
@@ -6,15 +6,40 @@
...
+
+    public static List<Attachment> getTestAttachments(String isContained) {
+        List<Attachment> attachments = new ArrayList<Attachment>();
+        attachments.add(buildTestAttchemnt("123", "test1.txt"));
+        attachments.add(buildTestAttchemnt(isContained, "test2.txt"));
+        attachments.add(buildTestAttchemnt("abc", "test3.txt"));
+        return attachments;
+    }
+
+    public static Attachment buildTestAttchemnt(String id, String name) {
```
markpit, line 36: attachment spelled wrong

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: Moved it to a separate method

mwestacott: Sorry, but that's not addressing my comment - the 'deleteFromS3' method doesn't require a company number or objection Id in order to do it's job. Therefore these things shouldn't be passed in. If we think that these values should be written to the log then maybe that should be job of the client method that calls 'deleteFromS3'.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/utils/Utils.java
mattch1: Corrected

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: Reworked this.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClientUnitTest.java
mattch1: camel cased

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -196,10 +194,47 @@ public void deleteAttachment(String requestId, String companyNumber, String obje
...
+
         attachments.remove(attachment);
 
         objection.setAttachments(attachments);
 
         objectionRepository.save(objection);
+
+    }
+
+    private void deleteFromS3(String requestId, String attachmentId, Map<String, Object>  logMap) throws ServiceException {
```
mwestacott, line 208: I've added another comment related to this, but I don't see the point in the logMap parameter. I don't think it's needed.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: To be honest I'm not sure what you have in mind.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -196,10 +194,47 @@ public void deleteAttachment(String requestId, String companyNumber, String obje
...
+        try {
+            FileTransferApiClientResponse response = fileTransferApiClient.delete(requestId, attachmentId);
+            if (response == null || response.getHttpStatus() == null) {
+                String message = String.format(DELETE_ERROR_MESSAGE_SHORT, attachmentId);
+                logger.infoContext(requestId, message, logMap);
```
mwestacott, line 213: Rather than 'infoContext' this and log messages below should use 'errorContext'.

mattch1: Well we have used info context in other places the reason being is that error context is used when we have an exception to pass in.

mwestacott: Ah ok. I thought there'd be an overloaded errorContext method that doesn't need an exception, but if there isn't then I guess infoContext is the one to use.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mwestacott: Think this is now resolved - logMap will only contain the attachmentId.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -49,6 +49,7 @@
     private static final String ERIC_AUTHORISED_USER = "ERIC-Authorised-User";
     private static final String OBJECTION_NOT_FOUND = "Objection not found";
     private static final String ATTACHMENT_NOT_FOUND = "Attachment not found";
+    public static final String COULD_NOT_DELETE = "Could not delete attachment";
```
mwestacott, line 52: Suggest getting 'ATTACHMENT' into this constant name for some context, like the others above. ATTACHMENT_NOT_DELETED perhaps?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -38,6 +40,8 @@
 
     private static final String OBJECTION_NOT_FOUND_MESSAGE = "Objection with id: %s, not found";
     private static final String ATTACHMENT_NOT_FOUND_MESSAGE = "Attachment with id: %s, not found";
+    private static final String DELETE_ERROR_MESSAGE = "Unable to delete attachment %s, status code %s";
```
mwestacott, line 43: Again, just some context in the name of these two constants - so prefixed with 'ATTACHMENT' and maybe remove 'ERROR' since the constants already defined above don't have this.

PR #31, opened 2020-07-21
OBJ-125 adding the get objection endpoint

SVarallo:
Jira ticket at: https://companieshouse.atlassian.net/browse/OBJ-125

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -22,6 +22,7 @@ import uk.gov.companieshouse.api.strikeoffobjections.common.LogConstants;
 import uk.gov.companieshouse.api.strikeoffobjections.exception.AttachmentNotFoundException;
 import uk.gov.companieshouse.api.strikeoffobjections.exception.ObjectionNotFoundException;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Attachment;
+import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
 import uk.gov.companieshouse.api.strikeoffobjections.model.patch.ObjectionPatch;
 import uk.gov.companieshouse.api.strikeoffobjections.model.response.AttachmentResponseDTO;
 import uk.gov.companieshouse.api.strikeoffobjections.model.response.ObjectionResponseDTO;
@@ -54,16 +55,19 @@ public class ObjectionController {
     private IObjectionService objectionService;
 
     private ApiLogger apiLogger;
+    private ObjectionMapper objectionMapper;
     private AttachmentMapper attachmentMapper;
 
     @Autowired
     public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
             IObjectionService objectionService,
             ApiLogger apiLogger,
+            ObjectionMapper objectionMapper,
             AttachmentMapper attachmentMapper) {
         this.responseEntityFactory = responseEntityFactory;
         this.objectionService = objectionService;
         this.apiLogger = apiLogger;
+        this.objectionMapper = objectionMapper;
         this.attachmentMapper = attachmentMapper;
     }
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -195,6 +199,45 @@ public ResponseEntity patchObjection(
...
+            return responseEntityFactory.createResponse(ServiceResult.found(responseDTO));
+        } catch (ObjectionNotFoundException e) {
+            apiLogger.errorContext(
+                    requestId,
+                    OBJECTION_NOT_FOUND,
+                    e,
+                    logMap
+            );
+
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
```
markpit, line 231: I would add another 'catch all' catch(Exception e) to handle anything other than not found exceptions. This would log the error and return a 500 error back to the user

markpit: it would go after the not found catch block

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
```diff
@@ -126,6 +130,29 @@ void  patchObjectionNotFoundExceptionTest() throws  Exception {
...
+        ResponseEntity<ChResponseBody<ObjectionResponseDTO>> response =
+                objectionController.getObjection(COMPANY_NUMBER, OBJECTION_ID, REQUEST_ID);
+        ChResponseBody<ObjectionResponseDTO> responseBody = response.getBody();
+        ObjectionResponseDTO responseDTO = responseBody.getSuccessBody();
+
+        verify(objectionMapper, times(1)).objectionEntityToObjectionResponseDTO(objection);
+
+        assertEquals(objection.getId(), responseDTO.getId());
+    }
+
```
markpit, line 155: you should have a test to handle an objection not found exception to make sure the controller returns the correct value.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mwestacott: Agree. We should add this to all the other controller end-point methods too, as it seems to be missing on most of them. Another sub-task and PR perhaps...

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
SVarallo: Added a test below to handle this

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
SVarallo: Added another catch block for the 500 error. We can maybe add this check into the other methods in a new story

ICoxCH: Could be worth creating a global restControllerAdvice exception handler, and just let the controllers throw the exceptions and the global handler will catch and handle them all, so we don't duplicate similar code throughout the controller.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
```diff
@@ -126,6 +130,38 @@ void  patchObjectionNotFoundExceptionTest() throws  Exception {
...
+    @Test
+    void getObjectionObjectionNotFoundTest() throws Exception {
```
markpit, line 157: now you've added the catch(exception) block, can you add a test to check that part works as well, so have another test for getObjectionUnexpectedExceptionThrown or something

PR #32, opened 2020-07-24
OBJ-155 Email model class

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/Email.java
```diff
@@ -0,0 +1,92 @@
...
+            createdAt = val;
+            return this;
+        }
+
+        public Email build() {
+            return new Email(this);
+        }
+    }
+    
+    private final String appId;
```
markpit, line 53: i'm not sure what appId is, could it be something like originatingAppId ( i assume)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/Email.java
```diff
@@ -0,0 +1,92 @@
+package uk.gov.companieshouse.api.strikeoffobjections.model.model;
+
+import java.util.Map;
+
+public class Email {
```
mattch1, line 5: Could we call this "EmailDocument" it will be consistent with what we have in other apps.

mattch1, line 5: this should implement serializable

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/Email.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public static final class Builder {
+        private String appId;
```
markpit, line 8: you could have a unit test to make sure the builder builds correctly :)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/Email.java
mattch1: I believe this is how it is set up within the chs infrastructure.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/Email.java
markpit: if this represents an email I'm happy with email

ICoxCH: https://github.com/companieshouse/promise-to-file-api/blob/master/src/main/java/uk/gov/companieshouse/api/promisetofile/model/email/EmailDocument.java#L3

You're right it probably should but the one in ptf doesn't

markpit: maybe get martin to decide!

mattch1: It looks like something somewhere is being too lenient.

ICoxCH: It represents the data that we put as a message on a kafka topic that the email sender constructs an email from. 
Not sure how best to describe it but I don't really know what "Email Document" means

markpit: EmailContent?

mattch1: It's just less confusing if someone looks at extension and ptf and wishes to see the equivalent in objections.

ICoxCH: I'll implement serializable here anyway

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/Email.java
ICoxCH: Looking at ptf, the kafka client needs to put a field called "app_id" on the message, but we can name this whatever.
So will change to originatingAppId so at a glance is clearer what is happening

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/Email.java
markpit: we should try and improve with each project otherwise we'd be stuck with how it was done in extensions forever

mattch1: In that case we should probably have something more descriptive of what it is for, something like "EmailWrapper" or "EmailContent".

ICoxCH: EmailContent it is

PR #33, opened 2020-07-27
OBJ-156 kafka client

mattch1:
Added Kafka client:
Avro serializer does not accept objects including Maps so had to convert the map for email data to a string.
DateFormatter added.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/FormatUtils.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/FormatUtils.java
@@ -5,13 +5,13 @@ import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 
 public class FormatUtils {
-    private static final String CREATE_AT_FORMAT = "dd MMM yyyy HH:mm:ss";
+    private static final String CREATED_AT_FORMAT = "dd MMM yyyy HH:mm:ss";
     private static final String DATE_FORMAT = "dd MMMM yyyy";
 
     private FormatUtils() {}
 
     public static String formatTimestamp(LocalDateTime timestamp) {
-        return timestamp.format(DateTimeFormatter.ofPattern(CREATE_AT_FORMAT));
+        return timestamp.format(DateTimeFormatter.ofPattern(CREATED_AT_FORMAT));
     }
 
     public static String formatDate(LocalDate date) {
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/FormatUtils.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/FormatUtils.java
@@ -1,0 +1,20 @@
+package uk.gov.companieshouse.api.strikeoffobjections.common;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class FormatUtils {
+    private static final String CREATE_AT_FORMAT = "dd MMM yyyy HH:mm:ss";
+    private static final String DATE_FORMAT = "dd MMMM yyyy";
+
+    private FormatUtils() {}
+
+    public static String formatTimestamp(LocalDateTime timestamp) {
+        return timestamp.format(DateTimeFormatter.ofPattern(CREATE_AT_FORMAT));
+    }
+
+    public static String formatDate(LocalDate date) {
+        return date.format(DateTimeFormatter.ofPattern(DATE_FORMAT));
+    }
+}
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/KafkaEmailClient.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/KafkaEmailClient.java
@@ -1,0 +1,57 @@
+package uk.gov.companieshouse.api.strikeoffobjections.email;
+
+import org.apache.avro.Schema;
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.strikeoffobjections.model.model.EmailContent;
+import uk.gov.companieshouse.kafka.message.Message;
+import uk.gov.companieshouse.kafka.producer.CHKafkaProducer;
+import uk.gov.companieshouse.service.ServiceException;
+
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.function.Supplier;
+
+@Component
+public class KafkaEmailClient {
+
+    @Value("${EMAIL_SEND_QUEUE_TOPIC}")
+    private String emailSendQueueTopic;
+
+    private CHKafkaProducer producer;
+    private AvroSerializer avroSerializer;
+    private Schema schema;
+
+    @Autowired
+    public KafkaEmailClient(CHKafkaProducer producer,
+                            AvroSerializer avroSerializer,
+                            Schema schema) {
+        this.producer = producer;
+        this.avroSerializer = avroSerializer;
+        this.schema = schema;
+    }
+
+    public void sendEmailToKafka(LocalDateTime currentTimestamp, EmailContent emailContent)
+            throws ServiceException {
+        try {
+            Message message = new Message();
+            byte[] serializedData = avroSerializer.serialize(emailContent, schema);
+            message.setValue(serializedData);
+            message.setTopic(emailSendQueueTopic);
+            message.setTimestamp(currentTimestamp.atZone(ZoneId.systemDefault()).toEpochSecond());
+            Future<RecordMetadata> future = producer.sendAndReturnFuture(message);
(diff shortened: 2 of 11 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/FormatUtils.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.companieshouse.api.strikeoffobjections.common;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+public class FormatUtils {
+    private static final String CREATE_AT_FORMAT = "dd MMM yyyy HH:mm:ss";
```
ICoxCH, line 8: should this be CREATED_AT_FORMAT?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/EmailContent.java
```diff
@@ -11,10 +14,12 @@
         private String originatingAppId;
         private String messageId;
         private String messageType;
-        private Map<String, Object> data;
```
ICoxCH, line 14: Small thing but I would suggest to keep the builder type as a Map, and only change it to a string on build()
In my head I wouldn't expect a "with" method to be transforming any data

src/test/java/email/KafkaEmailClientUnitTest.java
```diff
@@ -0,0 +1,105 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class KafkaEmailClientUnitTest {
+
+    private static final LocalDateTime CREATED_AT =
+            LocalDateTime.of(2019, 1, 1, 0, 0);
+    private static final Future<RecordMetadata> mockedFuture = Mockito.mock(Future.class);
+    private static final Future<RecordMetadata> faultyMmockedFuture = Mockito.mock(Future.class);
```
ICoxCH, line 38: private static final variables should be all caps, also this one has a typo

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/EmailContentTest.java
```diff
@@ -30,7 +34,7 @@ void emailBuilderTest() {
         assertEquals(emailContent.getOriginatingAppId(), ORIGINATING_APP_ID);
         assertEquals(emailContent.getMessageId(), MESSAGE_ID);
         assertEquals(emailContent.getMessageType(), MESSAGE_TYPE);
-        assertEquals(emailContent.getData(), DATA);
+        assertEquals(emailContent.getData(), objectMapper.writeValueAsString(DATA));
```
ICoxCH, line 37: this isn't a good test, as you are relying on the object mapper implementation to essentially test the object mapper implementation.
Would suggest defining a static final string as whatever the objectMapper produces and checking that value.

Also I realise I did this wrong, but would you mind reversing the argument order for all of the asserts in this class. as in (expectedValue, actualValue)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/KafkaEmailClient.java
```diff
@@ -0,0 +1,57 @@
...
+import uk.gov.companieshouse.kafka.message.Message;
+import uk.gov.companieshouse.kafka.producer.CHKafkaProducer;
+import uk.gov.companieshouse.service.ServiceException;
+
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.function.Supplier;
```
ICoxCH, line 18: Sonar says this import is unused

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/common/FormatUtils.java
mattch1: I prefer "CREATED" so changed this.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/EmailContent.java
mattch1: Moved this to build have had som make some changes to the mdoel to accommodate this.

src/test/java/email/KafkaEmailClientUnitTest.java
mattch1: They certainly should.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/EmailContentTest.java
mattch1: Replaced with a string for expected value and reversed paramaters in assert equals.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/KafkaEmailClient.java
mattch1: Removed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/EmailContent.java
```diff
@@ -15,6 +18,8 @@
         private String emailAddress;
         private String createdAt;
 
+        private final ObjectMapper objectMapper = new ObjectMapper();
```
markpit, line 21: I think it would be cleaner to do the objectMapping in the AvroSerializer and keep this pojo as a pojo rather than add any processing into it. That way I think it keeps this class in line with the 'Model' folder it lives in.

ICoxCH: To add to Marks comment, maybe we should change the createdAt field to contain a date object (LocalDateTime maybe?) and let the serializer deal with mapping that too, as it stands the time needs to be mapped before it is passed in

mattch1: Moved field conversion handling to AvroSerializer.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/EmailContent.java
```diff
@@ -15,6 +18,8 @@
...
+
         public Builder() {
```
mwestacott, line 19: You shouldn't need this empty argument constructor - Java provides it by default.

mattch1: Didn't add this, passed previous PR but removed empty constructor.

src/test/java/email/AvroSerializerUnitTest.java
```diff
@@ -0,0 +1,57 @@
...
+        assertTrue(result.contains(EMAIL_TEMPLATE_APP_ID));
+        assertTrue(result.contains(MESSAGE_ID));
+        assertTrue(result.contains(EMAIL_TEMPLATE_APP_ID));
+        assertTrue(result.contains(Utils.getDummyEmailData().get("to").toString()));
+        assertTrue(result.contains(Utils.getDummyEmailData().get("subject").toString()));
+        assertTrue(result.contains(Utils.getDummyEmailData().get("company_name").toString()));
+        assertTrue(result.contains(Utils.getDummyEmailData().get("company_number").toString()));
+        assertTrue(result.contains(Utils.getDummyEmailData().get("reason").toString()));
+        assertTrue(result.contains(RECIPIENT));
+        assertTrue(result.contains(FormatUtils.formatTimestamp(CREATED_AT)));
```
ICoxCH, line 54: This should be a hardcoded value as what is in there will be the result of the FormatUtil.formatTimestamp method, so if that method changes the output this test will still pass but it should fail

mattch1: Changed expected value.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/KafkaEmailClient.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.strikeoffobjections.model.model.EmailContent;
+import uk.gov.companieshouse.kafka.message.Message;
+import uk.gov.companieshouse.kafka.producer.CHKafkaProducer;
+import uk.gov.companieshouse.service.ServiceException;
+
+import java.io.IOException;
+import java.time.LocalDateTime;
```
ICoxCH, line 14: The sonar report is claiming this to be an unused import

mattch1: Thought I'd got them all removed this one.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/model/EmailContent.java
mwestacott: Yep, appreciate that. Thanks 👍 .

PR #34, opened 2020-07-27
OBJ-159 initial submit objection processing

markpit:
Adding an initial objection processor that is called when a patch containing status = SUBMITTED is sent in to the api.

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -85,7 +85,7 @@
         }
       },
       "patch": {
-        "summary": "Updates the strike-off objection identified by objectionId with the values provided",
+        "summary": "Updates the strike-off objection identified by objectionId with the values provided. If status set to SUBMITTED, this will trigger the Objection processing.",
         "produces": [
           "application/json"
         ],
@@ -107,6 +107,9 @@
           },
           "404": {
             "description": "objection not found"
+          },
+          "422": {
+            "description": "Unable to process objection"
           }
         }
       }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
@@ -54,7 +54,7 @@ public class ObjectionProcessor {
 
         Map<String, Object> logMap = new HashMap<>();
         logMap.put(LOG_OBJECTION_ID_KEY, objection.getId());
-        apiLogger.infoContext(httpRequestId, "Starting objection processing", logMap);
+        apiLogger.debugContext(httpRequestId, "Starting objection processing", logMap);
 
         validateObjectionStatus(objection, httpRequestId);
 
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
@@ -24,8 +24,9 @@ public class ObjectionProcessor {
 
     /**
      * Process the specified Objection
+     *
      * @param httpRequestId http request id used for logging
-     * @param objectionId id of the objection to process
+     * @param objectionId   id of the objection to process
      * @throws ObjectionNotFoundException if objectionId not found in database
      */
     public void process(String httpRequestId, String objectionId) throws ObjectionNotFoundException {
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
@@ -2,12 +2,19 @@ package uk.gov.companieshouse.api.strikeoffobjections.processor;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
+import uk.gov.companieshouse.api.strikeoffobjections.common.LogConstants;
 import uk.gov.companieshouse.api.strikeoffobjections.exception.ObjectionNotFoundException;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
+import uk.gov.companieshouse.api.strikeoffobjections.model.entity.ObjectionStatus;
 import uk.gov.companieshouse.api.strikeoffobjections.service.IObjectionService;
+
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Processes an Objection
+ *
  * Will only process Objection if status = SUBMITTED
  * Calls Chips to place stop against company
  * Sends email
@@ -15,11 +22,18 @@ import uk.gov.companieshouse.api.strikeoffobjections.service.IObjectionService;
 @Component
 public class ObjectionProcessor {
 
+    private static final String INVALID_START_STATUS_MSG =
+            "Objection %s has status %s. Cannot process unless status = SUBMITTED";
+    private static final String LOG_OBJECTION_ID_KEY = LogConstants.OBJECTION_ID.getValue();
+
     private IObjectionService objectionService;
+    private ApiLogger apiLogger;
 
     @Autowired
-    public ObjectionProcessor(IObjectionService objectionService) {
+    public ObjectionProcessor(IObjectionService objectionService,
+                              ApiLogger apiLogger) {
         this.objectionService = objectionService;
+        this.apiLogger = apiLogger;
     }
 
     /**
@@ -28,14 +42,13 @@ public class ObjectionProcessor {
      *
      * @param httpRequestId http request id used for logging
      * @param objectionId   id of the objection to process
-     * @throws ObjectionNotFoundException if objectionId not found in database
+     * @throws ObjectionNotFoundException      if objectionId not found in database
+     * @throws InvalidObjectionStatusException if the Objection is not currently in status SUBMITTED when this is called
      */
-    public void process(String httpRequestId, String objectionId) throws ObjectionNotFoundException {
+    public void process(String httpRequestId, String objectionId)
+            throws ObjectionNotFoundException, InvalidObjectionStatusException {
 
-        // TODO if status != submitted, throw an exception
-
-        // get objection
-        Objection objection = objectionService.getObjection(httpRequestId, objectionId);
+        Objection objection = getObjectionForProcessing(httpRequestId, objectionId);
 
         // TODO update status to processing
 
@@ -47,5 +60,26 @@ public class ObjectionProcessor {
 
         // TODO update status to processed
 
+    }
+
+    private Objection getObjectionForProcessing(String httpRequestId, String objectionId)
+            throws ObjectionNotFoundException, InvalidObjectionStatusException {
+
+        // get objection
+        Objection objection = objectionService.getObjection(httpRequestId, objectionId);
+
+        // if status not = submitted, throw exception
+        if (objection != null && ObjectionStatus.SUBMITTED != objection.getStatus()) {
+            InvalidObjectionStatusException statusException = new InvalidObjectionStatusException(
+                    String.format(INVALID_START_STATUS_MSG, objectionId, objection.getStatus()));
+
+            Map<String, Object> logMap = new HashMap<>();
+            logMap.put(LOG_OBJECTION_ID_KEY, objectionId);
+            apiLogger.errorContext(httpRequestId, statusException.getMessage(), statusException, logMap);
+
+            throw statusException;
+        }
+
+        return objection;
     }
 }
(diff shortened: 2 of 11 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -122,20 +122,20 @@ public Objection getObjection(String requestId, String objectionId) throws Objec
         String attachmentId = response.getFileId();
         if (StringUtils.isBlank(attachmentId)) {
             throw new ServiceException("No file id returned from file upload");
-        } else {
-            Attachment attachment = createAttachment(file, attachmentId);
-            Objection objection = objectionRepository.findById(objectionId).orElseThrow(
-                    () -> new ObjectionNotFoundException(String.format(OBJECTION_NOT_FOUND_MESSAGE, objectionId))
-            );
-            objection.addAttachment(attachment);
+        }
```
markpit, line 125: this is just formatting, removed the else block to reduce the levels of nesting (similar to the throw block above this one)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -219,14 +219,14 @@ private void deleteFromS3(String requestId, String attachmentId, Map<String, Obj
...
                 throw new ServiceException(message);
-            } else {
-                if (response.getHttpStatus().isError()) {
-                    String message = String.format(ATTACHMENT_NOT_DELETED,
-                            attachmentId, response.getHttpStatus());
-                    logger.infoContext(requestId, message, logMap);
-                    throw new ServiceException(message);
-                }
             }
+            if (response.getHttpStatus().isError()) {
```
markpit, line 223: just formatting, reducing levels of nesting (sonar complexity rating)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+     * @throws ObjectionNotFoundException if objectionId not found in database
+     */
+    public void process(String httpRequestId, String objectionId) throws ObjectionNotFoundException {
+
+        // if status != submitted, return (or throw ?)
+
+        // get objection
+        Objection objection = objectionService.getObjection(httpRequestId, objectionId);
+
+        // update status to processing
```
mattch1, line 39: Are there any tickets for these if so could we put the numbers on the comments? Also can they all be marked TODO so that we can see them in the IDE TODO tab

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: These could probably go into a separate method.

markpit: tried having a separate method but it didn't feel right checking the response from the call in a different method. I've refactored it, see if you think it is any better. if not I can revert it back to the original.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.strikeoffobjections.exception.ObjectionNotFoundException;
+import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
+import uk.gov.companieshouse.api.strikeoffobjections.service.IObjectionService;
+
+/**
+ * Processes an Objection
+ * Will only process Objection is status = SUBMITTED
```
mwestacott, line 11: 'if status'

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+     * Process the specified Objection
+     * Only processes if status is SUBMITTED
+     *
+     * @param httpRequestId http request id used for logging
+     * @param objectionId   id of the objection to process
+     * @throws ObjectionNotFoundException if objectionId not found in database
+     */
+    public void process(String httpRequestId, String objectionId) throws ObjectionNotFoundException {
+
+        // TODO if status != submitted, return (or throw ?)
```
mwestacott, line 35: Probably an error scenario so I'm thinking throw an exception. Better to fail noisily if we've got unexpected scenarios going on.

markpit: I can add that in

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: done

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: done

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -66,12 +65,12 @@ public ObjectionService(ObjectionRepository objectionRepository,
     }
 
     @Override
-    public String createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails) throws Exception{
+    public String createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails) {
```
ICoxCH, line 68: Does this not complain that it no longer matches the interface? As the interface createObjection throws Exception
Though if it doesn't then I agree we should remove, but i'd say remove from interface too?

spec/schema.json
```diff
@@ -85,7 +85,7 @@
         }
       },
       "patch": {
-        "summary": "Updates the strike-off objection identified by objectionId with the values provided",
+        "summary": "Updates the strike-off objection identified by objectionId with the values provided. If status set to SUBMITTED, this will trigger the Objection processing.",
```
mwestacott, line 88: 👍

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * Will only process Objection if status = SUBMITTED
+ * Calls Chips to place stop against company
+ * Sends email
+ */
+@Component
+public class ObjectionProcessor {
+
+    private static final String INVALID_START_STATUS_MSG =
+            "Objection %s has status %s. Cannot process unless status = SUBMITTED";
```
mwestacott, line 27: This message looks wrong, shouldn't it be 'unless status != SUBMITTED' based on logic check lower down?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    private Objection getObjectionForProcessing(String httpRequestId, String objectionId)
+            throws ObjectionNotFoundException, InvalidObjectionStatusException {
+
+        // get objection
+        Objection objection = objectionService.getObjection(httpRequestId, objectionId);
+
+        // if status not = submitted, throw exception
+        if (objection != null && ObjectionStatus.SUBMITTED != objection.getStatus()) {
```
mwestacott, line 77: I wonder if this check should be more specific and only allow processing when status is 'OPEN'. This would protect against 'PROCESSED' objections potentially being processed again, when we come to set that status...

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+
+        Objection objection = getObjectionForProcessing(httpRequestId, objectionId);
+
+        // TODO update status to processing
+
+        // TODO OBJ-139/OBJ-20 do chips sending
+
+        // TODO update status to chips sent
+
+        // TODO OBJ-113 do email sending
```
mwestacott, line 64: OBJ-113 is the story but what sub-task covers this work 'do email sending'? To be honest, I thought it was OBJ-159 that would actually stitch this together and so was expecting the call in this PR (once @ICoxCH 's changes are done).

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
markpit: The IDE highlights it as well in the interface as being too generic, i must have missed that one as well.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: I'm going to document the status flow that should help clear it up. In the meantime, I think it should be ok as is, the flow should go open->submitted->processing->chips_sent->processed.  We could have email_sent after chips_sent but nothing happens between email_sent and going to processed, so it seems better to go straight to processed which will be the finish state. I'm trying to draw up a finite state machine for it!

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: it should be ok, if status != submitted it will throw an InvalidObjectionStatusException and won't process it. My comment may have caused confusion, it used "not =" instead of "!="

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: I'm happy to do that work in here

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+
+        // TODO update status to processed
+
+    }
+
+    private Objection getObjectionForProcessing(String httpRequestId, String objectionId)
+            throws ObjectionNotFoundException, InvalidObjectionStatusException {
+
+        // get objection
+        Objection objection = objectionService.getObjection(httpRequestId, objectionId);
```
ICoxCH, line 74: Perhaps we should refactor the objectionService.patchObjection method to return the result of its repository.save() call. and pass the result into this method to save an uneeded db call

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
mwestacott: I guess this is ok, it's essentially repeating the logic check that's done in the controller PATCH method so in theory it will never fail/trigger, but no harm in repeating it.

The one thing that both the controller and this processor will allow though (I think) is for all the processing to be performed multiple times if the controller PATCH method is invoked multiple times with a status of 'SUBMITTED' - we'd further checks in the service method probably to implement that state machine your talking about and not allow invalid transitions via calls to the PATCH method. I actually think this is quite important with regard to robustness of the service. What do you think?

markpit: As discussed, created story OBJ-177

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: We could do that, or another alternative would be to have the ObjectionService.patchObjection call the processor (and pass the entity in to it) and the controller would just call the service? I think @mwestacott had an opinion previously about this as well.

mwestacott: How about the controller only calls the service and the service calls the processor, passing in the objection (which it should already have)? @markpit suggested this and I like the idea of keeping the controller nice and simple - delegating everything to the service.

markpit: ok, i can do that, should get rid of the extra db 'get' call then like Ian said.

PR #37, opened 2020-07-29
OBJ-175 added sdk client and service call

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/CompanyProfileService.java
```diff
@@ -1,20 +1,60 @@
...
+    public CompanyProfileService(ApiSdkClient apiSdkClient,
+                                 ApiLogger apiLogger) {
+        this.apiSdkClient = apiSdkClient;
+        this.apiLogger = apiLogger;
+    }
+
+    /**
+     * Calls the API SDK Manager Java library in order to retrieve the company (profile) details.
+     */
+    public CompanyProfileApi getCompanyProfile(String requestId, String companyNumber) throws ServiceException {
```
markpit, line 34: it's only a tiny thing, I've started to put the params in order of importance, might be better to have company number before requestId?

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/CompanyProfileServiceTest.java
```diff
@@ -0,0 +1,85 @@
...
+    @Mock
+    private CompanyGet companyGet;
+
+    @Mock
+    private ApiResponse<CompanyProfileApi> apiResponse;
+
+    @InjectMocks
+    private CompanyProfileService companyProfileService;
+
+    private CompanyProfileApi mockCompanyProfile;
```
markpit, line 52: could you rename mockCompanyProfile - maybe dummyCompanyProfile? mock makes me think it's a Mockito.mock

PR #38, opened 2020-07-30
Feature/obj 171 dissolution email in service

mattch1:
Added the internal emails to the service layer - made some adjustments seperated out common code in EmailService and created a new config class to allow for unit testing - move some common code here too.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
@@ -6,23 +6,27 @@ import org.springframework.stereotype.Component;
 @Component
 public class EmailConfig {
 
-    @Value("${EMAIL_SENDER_APP_ID}")
     private String originatingAppId;
-
-    @Value(("${EMAIL_SUBMITTED_EXTERNAL_TEMPLATE_MESSAGE_TYPE}"))
     private String submittedCustomerEmailType;
-
-    @Value(("${EMAIL_SUBMITTED_INTERNAL_TEMPLATE_MESSAGE_TYPE}"))
     private String submittedDissolutionTeamEmailType;
-
-    @Value(("${EMAIL_RECIPIENTS_CARDIFF}"))
     private String emailRecipientsCardiff;
-
-    @Value(("${EMAIL_RECIPIENTS_EDINBURGH}"))
     private String emailRecipientsEdinburgh;
-
-    @Value(("${EMAIL_RECIPIENTS_BELFAST}"))
     private String emailRecipientsBelfast;
+
+    public EmailConfig(
+            @Value("${EMAIL_SENDER_APP_ID}") String originatingAppId,
+            @Value(("${EMAIL_SUBMITTED_EXTERNAL_TEMPLATE_MESSAGE_TYPE}")) String submittedCustomerEmailType,
+            @Value(("${EMAIL_SUBMITTED_INTERNAL_TEMPLATE_MESSAGE_TYPE}")) String submittedDissolutionTeamEmailType,
+            @Value(("${EMAIL_RECIPIENTS_CARDIFF}")) String emailRecipientsCardiff,
+            @Value(("${EMAIL_RECIPIENTS_EDINBURGH}")) String emailRecipientsEdinburgh,
+            @Value(("${EMAIL_RECIPIENTS_BELFAST}"))String emailRecipientsBelfast) {
+        this.originatingAppId = originatingAppId;
+        this.submittedCustomerEmailType = submittedCustomerEmailType;
+        this.submittedDissolutionTeamEmailType = submittedDissolutionTeamEmailType;
+        this.emailRecipientsCardiff = emailRecipientsCardiff;
+        this.emailRecipientsEdinburgh = emailRecipientsEdinburgh;
+        this.emailRecipientsBelfast = emailRecipientsBelfast;
+    }
 
     public String getOriginatingAppId() {
         return originatingAppId;
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
@@ -6,6 +6,7 @@ import org.springframework.stereotype.Component;
 @Component
 public class EmailConfig {
 
+    private String emailSubject;
     private String originatingAppId;
     private String submittedCustomerEmailType;
     private String submittedDissolutionTeamEmailType;
@@ -14,18 +15,24 @@ public class EmailConfig {
     private String emailRecipientsBelfast;
 
     public EmailConfig(
+            @Value("${EMAIL_SUBJECT}") String emailSubject,
             @Value("${EMAIL_SENDER_APP_ID}") String originatingAppId,
             @Value(("${EMAIL_SUBMITTED_EXTERNAL_TEMPLATE_MESSAGE_TYPE}")) String submittedCustomerEmailType,
             @Value(("${EMAIL_SUBMITTED_INTERNAL_TEMPLATE_MESSAGE_TYPE}")) String submittedDissolutionTeamEmailType,
             @Value(("${EMAIL_RECIPIENTS_CARDIFF}")) String emailRecipientsCardiff,
             @Value(("${EMAIL_RECIPIENTS_EDINBURGH}")) String emailRecipientsEdinburgh,
             @Value(("${EMAIL_RECIPIENTS_BELFAST}"))String emailRecipientsBelfast) {
+        this.emailSubject = emailSubject;
         this.originatingAppId = originatingAppId;
         this.submittedCustomerEmailType = submittedCustomerEmailType;
         this.submittedDissolutionTeamEmailType = submittedDissolutionTeamEmailType;
         this.emailRecipientsCardiff = emailRecipientsCardiff;
         this.emailRecipientsEdinburgh = emailRecipientsEdinburgh;
         this.emailRecipientsBelfast = emailRecipientsBelfast;
+    }
+
+    public String getEmailSubject() {
+        return emailSubject;
     }
 
     public String getOriginatingAppId() {
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
@@ -1,0 +1,50 @@
+package uk.gov.companieshouse.api.strikeoffobjections.email;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
+public class EmailConfig {
+
+    @Value("${EMAIL_SENDER_APP_ID}")
+    private String originatingAppId;
+
+    @Value(("${EMAIL_SUBMITTED_EXTERNAL_TEMPLATE_MESSAGE_TYPE}"))
+    private String submittedCustomerEmailType;
+
+    @Value(("${EMAIL_SUBMITTED_INTERNAL_TEMPLATE_MESSAGE_TYPE}"))
+    private String submittedDissolutionTeamEmailType;
+
+    @Value(("${EMAIL_RECIPIENTS_CARDIFF}"))
+    private String emailRecipientsCardiff;
+
+    @Value(("${EMAIL_RECIPIENTS_EDINBURGH}"))
+    private String emailRecipientsEdinburgh;
+
+    @Value(("${EMAIL_RECIPIENTS_BELFAST}"))
+    private String emailRecipientsBelfast;
+
+    public String getOriginatingAppId() {
+        return originatingAppId;
+    }
+
+    public String getSubmittedCustomerEmailType() {
+        return submittedCustomerEmailType;
+    }
+
+    public String getSubmittedDissolutionTeamEmailType() {
+        return submittedDissolutionTeamEmailType;
+    }
+
+    public String getEmailRecipientsCardiff() {
+        return emailRecipientsCardiff;
+    }
+
+    public String getEmailRecipientsEdinburgh() {
+        return emailRecipientsEdinburgh;
+    }
+
+    public String getEmailRecipientsBelfast() {
+        return emailRecipientsBelfast;
+    }
+}
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
@@ -10,4 +10,11 @@ public interface IEmailService {
             String companyNumber,
             Objection objection
     ) throws ServiceException;
+
+    void sendObjectionSubmittedDissolutionTeamEmail(
+            String requestId,
+            String ericAuthorisedUser,
+            String companyNumber,
+            Objection objection
+    ) throws ServiceException;
 }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
@@ -5,16 +5,16 @@ import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
 import uk.gov.companieshouse.service.ServiceException;
 
 public interface IEmailService {
+
     void sendObjectionSubmittedCustomerEmail(
-            String requestId,
-            String ericAuthorisedUser,
-            CompanyProfileApi companyProfile,
-            Objection objection
+            Objection objection,
+            String companyName,
+            String requestId
     ) throws ServiceException;
 
     void sendObjectionSubmittedDissolutionTeamEmail(
-            String requestId,
             CompanyProfileApi companyProfile,
-            Objection objection
+            Objection objection,
+            String requestId
     ) throws ServiceException;
 }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
@@ -1,6 +1,5 @@
 package uk.gov.companieshouse.api.strikeoffobjections.service;
 
-import uk.gov.companieshouse.api.model.company.CompanyProfileApi;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
 import uk.gov.companieshouse.service.ServiceException;
 
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
@@ -1,5 +1,6 @@
 package uk.gov.companieshouse.api.strikeoffobjections.service;
 
+import uk.gov.companieshouse.api.model.company.CompanyProfileApi;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
 import uk.gov.companieshouse.service.ServiceException;
 
@@ -7,13 +8,13 @@ public interface IEmailService {
     void sendObjectionSubmittedCustomerEmail(
             String requestId,
             String ericAuthorisedUser,
-            String companyNumber,
+            CompanyProfileApi companyProfile,
             Objection objection
     ) throws ServiceException;
 
     void sendObjectionSubmittedDissolutionTeamEmail(
             String requestId,
-            String companyNumber,
+            CompanyProfileApi companyProfile,
             Objection objection
     ) throws ServiceException;
 }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
@@ -13,7 +13,8 @@ public interface IEmailService {
     ) throws ServiceException;
 
     void sendObjectionSubmittedDissolutionTeamEmail(
-            CompanyProfileApi companyProfile,
+            String companyName,
+            String jurisdiction,
             Objection objection,
             String requestId
     ) throws ServiceException;
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
@@ -13,7 +13,6 @@ public interface IEmailService {
 
     void sendObjectionSubmittedDissolutionTeamEmail(
             String requestId,
-            String ericAuthorisedUser,
             String companyNumber,
             Objection objection
     ) throws ServiceException;
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
@@ -1,7 +1,6 @@
 package uk.gov.companieshouse.api.strikeoffobjections.service.impl;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import uk.gov.companieshouse.api.model.company.CompanyProfileApi;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
@@ -50,18 +49,18 @@ public class EmailService implements IEmailService {
     public void sendObjectionSubmittedCustomerEmail (
             String requestId,
             String ericAuthorisedUser,
-            String companyNumber,
+            CompanyProfileApi companyProfile,
             Objection objection
     ) throws ServiceException {
 
-        CompanyProfileApi companyProfile = companyProfileService.getCompanyProfile(companyNumber, requestId);
-
-        String companyName = companyProfile.getCompanyName();
-        Map<String, Object> data = constructEmailDataMap(companyName, companyNumber, objection);
+        Map<String, Object> data = constructEmailDataMap(
+                companyProfile.getCompanyName(),
+                companyProfile.getCompanyNumber(),
+                objection);
 
         String emailAddress = ericHeaderParser.getEmailAddress(ericAuthorisedUser);
         EmailContent emailContent = constructEmailContent(EmailType.CUSTOMER,
-                requestId, emailAddress, data);
+                emailAddress, data);
 
         logger.debugContext(requestId, "Calling Kafka client to send customer email");
         kafkaEmailClient.sendEmailToKafka(emailContent);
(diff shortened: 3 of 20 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -55,29 +53,83 @@ public void sendObjectionSubmittedCustomerEmail (
...
-        EmailContent emailContent = new EmailContent.Builder()
-                .withOriginatingAppId(originatingAppId)
+        CompanyProfileApi companyProfile = companyProfileService.getCompanyProfile(companyNumber, requestId);
+        String companyName = companyProfile.getCompanyName();
+        Map<String, Object> data = constructEmailDataMap(companyName, companyNumber, objection);
+
+        for (String emailAddress : getDissolutionTeamRecipients(companyProfile.getJurisdiction())) {
+            EmailContent emailContent = constructEmailContent(EmailType.DISSOLUTION_TEAM,
+                    requestId, emailAddress, data);
+            logger.debugContext(requestId, String.format("Calling Kafka client to send dissolution team emailto %s",
```
ICoxCH, line 85: missing space in emailto

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -55,29 +53,83 @@ public void sendObjectionSubmittedCustomerEmail (
...
+    @Override
+    public void sendObjectionSubmittedDissolutionTeamEmail(
+            String requestId,
+            String companyNumber,
+            Objection objection
+    ) throws ServiceException {
 
-        EmailContent emailContent = new EmailContent.Builder()
-                .withOriginatingAppId(originatingAppId)
+        CompanyProfileApi companyProfile = companyProfileService.getCompanyProfile(companyNumber, requestId);
```
ICoxCH, line 78: I think the companyProfile should be called in the ObjectionService and passed into the EmailService, as it is we do the same profile call twice, once for external once for internal.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
```diff
@@ -0,0 +1,50 @@
+package uk.gov.companieshouse.api.strikeoffobjections.email;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
+public class EmailConfig {
+
+    @Value("${EMAIL_SENDER_APP_ID}")
```
ICoxCH, line 9: This could be up for discussion, but may be worth Injecting these values into the constructor, and making the variables final.
This way Spring will refuse to start if the config isn't set
Something like

```
private final String originatingAppId;

public EmailConfig(@Value("${EMAIL_SENDER_APP_ID}") String orginiatingAppId) {
this.originatingAppId = originatingAppId;
}
```

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -55,29 +53,83 @@ public void sendObjectionSubmittedCustomerEmail (
...
+                    requestId, emailAddress, data);
+            logger.debugContext(requestId, String.format("Calling Kafka client to send dissolution team emailto %s",
+                    emailAddress));
+            kafkaEmailClient.sendEmailToKafka(emailContent);
+            logger.debugContext(requestId, "Successfully called Kafka client");
+        }
+    }
+
+    private EmailContent constructEmailContent(EmailType emailType,
+                                               String requestId,
```
ICoxCH, line 93: Sonar says the requestId is not used, I don't think there is anything here worth logging so would recommend removing the argument.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -55,29 +53,83 @@ public void sendObjectionSubmittedCustomerEmail (
...
+        data.put("company_number", companyNumber);
+        data.put("objection_id", objection.getId());
+        data.put("reason", objection.getReason());
+        data.put("attachments", objection.getAttachments());
+        return data;
+    }
+
+    protected String[] getDissolutionTeamRecipients(String jurisdiction) {
+        switch(jurisdiction) {
+            case "england":
```
ICoxCH, line 123: Personally i'd prefer not to let england and wales fall through into england-wales and explicitly define their behaviour (even though its the same as england-wales) just to improve code readability

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -38,6 +39,14 @@
     private static final String AUTH_USER = EMAIL + "; forename=demoForename; surname=demoSurname";
     private static final String REASON = "THIS IS A REASON";
 
+    private static final String EMAIL_RECIPIENTS_CARDIFF_TEST =
+            "COTIndividual@companieshouse.gov.uk,COTExttest@companieshouse.gov.uk,dissolution@companieshouse.gov.uk";
```
ICoxCH, line 43: Would suggest using clearly fake data here as this code is on a public repo. I don't think anything here is unreasonable but probably best something like "Fake1@Cardiff.com"

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -55,29 +53,83 @@ public void sendObjectionSubmittedCustomerEmail (
...
-        data.put("attachments", objection.getAttachments());
+        logger.debugContext(requestId, "Calling Kafka client to send customer email");
+        kafkaEmailClient.sendEmailToKafka(emailContent);
+        logger.debugContext(requestId, "Successfully called Kafka client");
+    }
+
+    @Override
+    public void sendObjectionSubmittedDissolutionTeamEmail(
+            String requestId,
+            String companyNumber,
```
markpit, line 74: you can remove company number from the input params, it is on the Objection

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -38,6 +39,14 @@
     private static final String AUTH_USER = EMAIL + "; forename=demoForename; surname=demoSurname";
     private static final String REASON = "THIS IS A REASON";
```
markpit, line 40: REASON is now unused

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -1,12 +1,14 @@
 package uk.gov.companieshouse.api.strikeoffobjections.service.impl;
 
+import org.junit.jupiter.api.BeforeEach;
```
markpit, line 3: BeforeEach is showing up as unused in IDE

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -96,6 +96,103 @@ void sendObjectionSubmittedCustomerEmail() throws ServiceException {
         assertTrue(data.containsValue("Company: " + COMPANY_NUMBER));
         assertTrue(data.containsValue(OBJECTION_ID));
         assertTrue(data.containsValue(attachments));
+    }
+
+    @Test
+    void sendObjectionSubmittedDissolutonEmailsWalesJurisdiction() throws ServiceException {
```
markpit, line 102: Dissolution spelling mistake

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -96,6 +96,103 @@ void sendObjectionSubmittedCustomerEmail() throws ServiceException {
...
+        verify(kafkaEmailClient, times(3)).sendEmailToKafka(emailContentArgumentCaptor.capture());
+    }
+
+    @Test
+    void sendObjectionSubmittedDissolutonEmailsScotlandJurisdiction() throws ServiceException {
+        when(dateTimeSupplier.get()).thenReturn(LOCAL_DATE_TIME);
+        when(companyProfileService.getCompanyProfile(COMPANY_NUMBER, REQUEST_ID))
+                .thenReturn(Utils.getDummyCompanyProfile(COMPANY_NUMBER, "scotland"));
+        when(config.getEmailRecipientsEdinburgh()).thenReturn(EMAIL_RECIPIENTS_EDINBURGH_TEST);
+        List<Attachment> attachments = Utils.getTestAttachments();
```
markpit, line 127: attachments var is unused

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -96,6 +96,103 @@ void sendObjectionSubmittedCustomerEmail() throws ServiceException {
...
+        verify(kafkaEmailClient, times(2)).sendEmailToKafka(emailContentArgumentCaptor.capture());
+    }
+
+    @Test
+    void sendObjectionSubmittedDissolutonEmailsNIJurisdiction() throws ServiceException {
+        when(dateTimeSupplier.get()).thenReturn(LOCAL_DATE_TIME);
+        when(companyProfileService.getCompanyProfile(COMPANY_NUMBER, REQUEST_ID))
+                .thenReturn(Utils.getDummyCompanyProfile(COMPANY_NUMBER, "northern-ireland"));
+        when(config.getEmailRecipientsBelfast()).thenReturn(EMAIL_RECIPIENTS_BELFAST_TEST);
+        List<Attachment> attachments = Utils.getTestAttachments();
```
markpit, line 147: attachments unused

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: Now passing company profile

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -21,12 +22,7 @@
 @Service
 public class EmailService implements IEmailService {
 
-    @Value("${EMAIL_SENDER_APP_ID}")
```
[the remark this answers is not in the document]
mattch1: Removed unused import

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/EmailConfig.java
mattch1: Put into a constructor

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -55,29 +53,83 @@ public void sendObjectionSubmittedCustomerEmail (
...
+            logger.debugContext(requestId, "Successfully called Kafka client");
+        }
+    }
+
+    private EmailContent constructEmailContent(EmailType emailType,
+                                               String requestId,
+                                               String emailAddress,
+                                               Map<String, Object> data) {
+
+        String typeOfEmail = (emailType == EmailType.CUSTOMER)? emailConfig.getSubmittedCustomerEmailType()
```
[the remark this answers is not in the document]
mattch1: Not sure about this.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: Removed unused.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: all cases now with return statements

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Renamed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: Removed

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Removed

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Removed

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Corrected.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Removed

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Removed

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -96,6 +88,96 @@ void sendObjectionSubmittedCustomerEmail() throws ServiceException {
...
+        Utils.getTestAttachments().forEach(objection::addAttachment);
+
+        emailService.sendObjectionSubmittedDissolutionTeamEmail(
+                REQUEST_ID,
+                Utils.getDummyCompanyProfile(COMPANY_NUMBER, "wales"),
+                objection
+        );
+
+        ArgumentCaptor<EmailContent> emailContentArgumentCaptor = ArgumentCaptor.forClass(EmailContent.class);
+        verify(kafkaEmailClient, times(3)).sendEmailToKafka(emailContentArgumentCaptor.capture());
```
markpit, line 107: The captured arguments aren't being asserted against so it doesn't test if the method is behaving correctly. For this test we need to check that england/wales emails are sent. To do that we'd need to check the contents of the arguments captured to see if they contain the correct values (email addresses etc..)

For multiple calls, this is an example from the Mockito javadoc:
`ArgumentCaptor<Person> peopleCaptor = ArgumentCaptor.forClass(Person.class);
verify(mock, times(2)).doSomething(peopleCaptor.capture());

List<Person> capturedPeople = peopleCaptor.getAllValues();
assertEquals("John", capturedPeople.get(0).getName());
assertEquals("Jane", capturedPeople.get(1).getName());`

more examples and docs - https://javadoc.io/static/org.mockito/mockito-core/3.4.6/org/mockito/ArgumentCaptor.html

This applies to the other tests as well.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
markpit: I think Ian meant to pass a CompanyProfile rather than the Api. But if we only need the company name (company number is already in the Objection) then we can just have the company name as a String input param to this method.

markpit: Sorry, CompanyProfileApi is a bad name! (i know it's not ours) ok I can see it's a pojo, not the api. But I think we only need the company name instead of the whole profile.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -96,6 +88,96 @@ void sendObjectionSubmittedCustomerEmail() throws ServiceException {
...
+                Utils.getDummyCompanyProfile(COMPANY_NUMBER, "northern-ireland"),
+                objection
+        );
+
+        ArgumentCaptor<EmailContent> emailContentArgumentCaptor = ArgumentCaptor.forClass(EmailContent.class);
+        verify(kafkaEmailClient, times(2)).sendEmailToKafka(emailContentArgumentCaptor.capture());
+    }
+
+    @Test
+    public void testRegionalEmailAddresses() {
```
markpit, line 145: you can remove 'public' to keep it consistent with the other tests in this class

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -1,10 +1,10 @@
...
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import uk.gov.companieshouse.api.model.company.CompanyProfileApi;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
+import uk.gov.companieshouse.api.strikeoffobjections.email.EmailConfig;
 import uk.gov.companieshouse.api.strikeoffobjections.email.KafkaEmailClient;
 import uk.gov.companieshouse.api.strikeoffobjections.model.email.EmailContent;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
```
markpit, line 10: ICompanyProfileService is now unused after you last commit

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -7,6 +7,7 @@
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
+import uk.gov.companieshouse.api.strikeoffobjections.email.EmailConfig;
 import uk.gov.companieshouse.api.strikeoffobjections.email.KafkaEmailClient;
 import uk.gov.companieshouse.api.strikeoffobjections.model.email.EmailContent;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Attachment;
```
markpit, line 13: ICompanyProfileService is unused import

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Updated testing.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Removed.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
mattch1: Removed.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: Removed.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: Changed CompanyProfile for companyName for customer emails.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
markpit, line 10: Unused import ICompanyProfileService

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -21,66 +20,113 @@
...
+                emailAddress, data);
+
+        logger.debugContext(requestId, "Calling Kafka client to send customer email");
+        kafkaEmailClient.sendEmailToKafka(emailContent);
+        logger.debugContext(requestId, "Successfully called Kafka client");
+    }
+
+    @Override
+    public void sendObjectionSubmittedDissolutionTeamEmail(
+            CompanyProfileApi companyProfile,
```
markpit, line 64: Looks like we only need the jurisdiction and companyName out of companyProfile, I think it would be better to just pass those 2 strings in as args rather than a CompanyProfileApi, it would be consistent with the customer email args as well.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessorTest.java
```diff
@@ -20,33 +21,46 @@
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static uk.gov.companieshouse.api.strikeoffobjections.common.LogConstants.COMPANY_NUMBER;
```
markpit, line 25: Unused import COMPANY_NUMBER

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -8,6 +8,7 @@
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.test.util.ReflectionTestUtils;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
+import uk.gov.companieshouse.api.strikeoffobjections.email.EmailConfig;
```
markpit, line 11: Multiple unused imports

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -35,18 +35,23 @@
     private static final String EMAIL_SUBJECT = "{{ COMPANY_NUMBER }}: email sent";
```
markpit, line 35: EMAIL_SUBJECT is unused

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -59,30 +64,18 @@
 
     @Test
     void sendObjectionSubmittedCustomerEmail() throws ServiceException {
-        ReflectionTestUtils.setField(emailService, "emailSubject", EMAIL_SUBJECT);
-        when(companyProfileService.getCompanyProfile(COMPANY_NUMBER, REQUEST_ID))
-                .thenReturn(Utils.getDummyCompanyProfile(COMPANY_NUMBER));
+        when(config.getEmailSubject()).thenReturn(FORMATTED_EMAIL_SUBJECT);
+        when(config.getEmailSubject()).thenReturn(FORMATTED_EMAIL_SUBJECT);
```
markpit, line 68: duplicate mocking

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/utils/Utils.java
```diff
@@ -10,6 +10,7 @@
 import uk.gov.companieshouse.api.model.company.CompanyProfileApi;
```
markpit, line 10: Unused import JsonProcessingException

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -1,14 +1,13 @@
 package uk.gov.companieshouse.api.strikeoffobjections.service.impl;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 import uk.gov.companieshouse.api.model.company.CompanyProfileApi;
```
markpit, line 5: Another unused import

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IEmailService.java
```diff
@@ -1,11 +1,20 @@
 package uk.gov.companieshouse.api.strikeoffobjections.service;
 
+import uk.gov.companieshouse.api.model.company.CompanyProfileApi;
```
markpit, line 3: Unused import

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: I have no idea howe these keep showing up I am deleting them all and then there are more of them.  I have just cleaned up EmailsService - the ide was showng none.

markpit: Everyone should have the SonarLint plugin installed (in intelliJ), there should be an equivalent for Eclipse

mwestacott: Yep. That plug-in works in Eclipse too. I have it installed. If you are using Eclipse @mattch1 , SHIFT-OPTION-O cleans up your imports.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -23,69 +21,118 @@
...
+            case "wales":
+                return emailConfig.getEmailRecipientsCardiff().split(",");
+            case "england-wales":
+                return emailConfig.getEmailRecipientsCardiff().split(",");
+            case "scotland":
+                return emailConfig.getEmailRecipientsEdinburgh().split(",");
+            case "northern-ireland":
+                return emailConfig.getEmailRecipientsBelfast().split(",");
+            default:
+                return emailConfig.getEmailRecipientsCardiff().split(",");
```
mwestacott, line 135: Are we sure this isn't a (runtime exception) error scenario, if jurisdiction doesn't match what we expect?

mattch1: No we dealt with this before - Cardiff is the default.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -23,69 +21,118 @@
...
-        kafkaEmailClient.sendEmailToKafka(emailContent);
-        logger.debugContext(requestId, "Successfully called Kafka client");
+    protected String[] getDissolutionTeamRecipients(String jurisdiction) {
+        switch(jurisdiction) {
+            case "england":
+                return emailConfig.getEmailRecipientsCardiff().split(",");
+            case "wales":
+                return emailConfig.getEmailRecipientsCardiff().split(",");
+            case "england-wales":
+                return emailConfig.getEmailRecipientsCardiff().split(",");
```
mwestacott, line 129: Appreciate @ICoxCH 's comment above and I haven't seen original value of code, but not using something like:

case "england": case "wales": case "england-wales":
   return emailConfig.getEmailRecipientsCardiff().split(",");
...

kind of negates the use of switch/case - it's intended to improve readability and cut down on repeated code.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: Yeh I know - I had a comment above about this saying the opposite - I think there are multiple points of view here.

mwestacott: Ok. Something like that always worth a comment in the code I think, but won't hold up my approval.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
mattch1: Yeh I know - I had a comment above about this saying the opposite - I think there are multiple points of view here.

PR #40, opened 2020-07-30
Added subject and to fields to email data

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -21,6 +21,8 @@
 @Service
 public class EmailService implements IEmailService {
 
+    private static final String EXTERNAL_EMAIL_SUBJECT = "%s: Objection Application Submitted";
```
mwestacott, line 24: I think you need to use the EMAIL_SUBJECT config value we defined and replace the placeholder, unless we think that approach is no longer needed.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailServiceTest.java
```diff
@@ -95,6 +95,8 @@ void sendObjectionSubmittedCustomerEmail() throws ServiceException {
         assertTrue(data.containsValue("Company: " + COMPANY_NUMBER));
         assertTrue(data.containsValue(OBJECTION_ID));
         assertTrue(data.containsValue(attachments));
+        assertTrue(data.containsValue(EMAIL));
+        assertTrue(data.containsKey("subject"));
```
mwestacott, line 99: I think you need a test for the value also, checking that the formatting/string replacement has worked as expected.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
ICoxCH: In my haste completely forgot this value was in configs, using that now

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -47,6 +48,13 @@ public EmailService(
         this.dateTimeSupplier = dateTimeSupplier;
     }
 
+    /**
```
mwestacott, line 51: Rather than add this in, use ReflectionTestUtils.setField() method in your test to set the subject.

PR #41, opened 2020-07-31
OBJ-180 Send a submitted date to the notification sender

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -69,6 +73,7 @@ public void sendObjectionSubmittedCustomerEmail(
         data.put("objection_id", objection.getId());
         data.put("reason", objection.getReason());
         data.put("attachments", objection.getAttachments());
+        data.put("date", FormatUtils.formatDate(submittedOn));
```
SVarallo, line 76: Is there an order to how these are added? Wouldn't it make logical sense to put them in the order they appear in the email?

ICoxCH: From a technical perspective I don't think java Hashmaps have an ordering so it doesn't matter.
From a code readability perspective, I shall re order them

PR #43, opened 2020-07-31
OBJ-172 adding send dissolution team email call to ObjectionProcessor

SVarallo:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -74,7 +74,18 @@ public void process(Objection objection, String httpRequestId)
 
         // TODO update status to CHIPS_SENT
 
-        // TODO OBJ-172 send internal email
+        // send internal email
+        try {
+            CompanyProfileApi companyProfile = this.companyProfileService.getCompanyProfile(objection.getCompanyNumber(), httpRequestId);
```
mwestacott, line 79: Now that you retrieve company profile here, you can remove the duplicate call lower down (line 94).

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -74,7 +74,18 @@ public void process(Objection objection, String httpRequestId)
...
+        } catch (Exception e) {
+            logMap = new HashMap<>();
+            logMap.put(LOG_OBJECTION_ID_KEY, objection.getId());
+            apiLogger.errorContext(httpRequestId, "Error sending dissolution team email", e,
+                    logMap);
+
+            throw e;
+        }
 
         // TODO update status to INTERNAL_EMAIL_SENT
```
mwestacott, line 90: Not this PR, but just wondering if we have a JIRA story/ticket for updating all these statuses?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
mattch1, line 79: Could put this line outside the try block as it is common to both emails

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -74,7 +74,18 @@ public void process(Objection objection, String httpRequestId)
 
         // TODO update status to CHIPS_SENT
 
-        // TODO OBJ-172 send internal email
+        // send internal email
+        try {
+            CompanyProfileApi companyProfile = this.companyProfileService.getCompanyProfile(objection.getCompanyNumber(), httpRequestId);
+            emailService.sendObjectionSubmittedDissolutionTeamEmail(companyProfile.getCompanyName(), companyProfile.getJurisdiction(), objection, httpRequestId);
+        } catch (Exception e) {
+            logMap = new HashMap<>();
```
mattch1, line 82: Then both email calls could go in the same try.. catch block then add log info the email being sent - customer /dissolution and then have a generic exception message. Alternatively set a string varaible for the email being sent and make that aprt of the exception message so that it is still tailored.

mwestacott: Yep, agree with @mattch1 in that we could have one try/catch block. As long as the exception stack trace is present in the logged error message (it is/should be) then we have the line number and we'll know what e-mail failed. This is kind of the intention of Java try/catch blocks - code for the sunny days scenario and deal with all the error cases below.

SVarallo: I'm not sure about why you would cause a need for someone to look at the code by having generic logs when you could just be specific, as these logs will be for the support staff, not us. It was Mark who originally suggested keeping these try/catch blocks separate to make this easier. If you guys are sure though I don't mind changing it!

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: Actually, you could make a private method called sendEmails(...) and have the company profile lookup and email sending / catching in there
And get the company profile once only, before sending any emails

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: ok i'm happy to go with a generic catch fr both calls, but keep the TODO INTERNAL_EMAIL_SENT status change in between them, we'll need to do that in future

markpit: Thinking about this some more, I'd like to keep them separate as this will give us the ability to set separate error status' indicating what type of email failed and will also allow support team to re-run them independently in future

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
markpit: OBJ-177, I've started doing some work on it

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -109,4 +99,34 @@ private void validateObjectionStatus(Objection objection, String httpRequestId)
...
+            emailService.sendObjectionSubmittedCustomerEmail(objection, companyProfile.getCompanyName(), httpRequestId);
+        } catch (Exception e) {
+            Map<String, Object> logMap = new HashMap<>();
+            logMap.put(LOG_OBJECTION_ID_KEY, objection.getId());
+            apiLogger.errorContext(httpRequestId, "Error sending customer email", e, logMap);
+
+            throw e;
+        }
+
+        // TODO update status to processed
```
markpit, line 129: can you move the todo back to the main public method, so it is the last line of the method

PR #44, opened 2020-08-04
OBJ-184 Send email field to notification api

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -70,10 +70,12 @@ public void sendObjectionSubmittedDissolutionTeamEmail(
 
 
         for (String emailAddress : getDissolutionTeamRecipients(jurisdiction)) {
-            Map<String, Object> data = constructEmailDataMap(
+            Map<String, Object> data = constructCommonEmailMap(
                     companyName,
                     objection,
                     emailAddress);
+
+            data.put("email", objection.getCreatedBy().getEmail());
```
markpit, line 78: can't we just use the 'to' field from data?  constructCommonEmailMap takes an email and puts it in the 'to' key.

markpit: just realised, customer email is different to internal 'to' address

markpit: could we not call it customerEmail if that's what it's for or would that mean changing the templates?

ICoxCH: The "to" field is where the email gets sent,
this field, is displayed in the internal email so they are different values

markpit: sorry , this is really confusing. emailAddress in the for loop should be renamed internalEmailAddress or dissolutionTeamEmailAddress and the data.put should happen inside the constructCommonEmailMap and possibly the key should be called customerEmail / externalEmail

markpit: ok, so "email" lives outside common... as it's only for internal emails. Just finiding it hard to deferentiate between emailAddress/to/email

ICoxCH: if we are to change to key, we would also need to change the template (not against it it is all a bit confusing)

markpit: ok leave as is

PR #45, opened 2020-08-04
Remove spaces from config entry

ICoxCH:


mwestacott:
Would be good to get a JIRA ticket number referencing the bug/story into the commit message.

ICoxCH:
> Would be good to get a JIRA ticket number referencing the bug/story into the commit message.

there isn't one, I can make one I guess.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/EmailService.java
```diff
@@ -123,11 +123,11 @@ private EmailContent constructEmailContent(EmailType emailType,
     protected String[] getDissolutionTeamRecipients(String jurisdiction) {
         switch(jurisdiction) {
             case "scotland":
-                return emailConfig.getEmailRecipientsEdinburgh().split(",");
+                return emailConfig.getEmailRecipientsEdinburgh().replace(" ", "").split(",");
```
mwestacott, line 126: Would be nice to have a local private 'removeSpaces()' method or similar that takes the list of email addresses and does the replace(" ", "") rather than seeing it repeated here. Minor, so will approve. See what others think.

PR #46, opened 2020-08-10
Feature/obj 198 stub transfer file client download

mattch1:
Stub method for downloads and stub test.


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -153,4 +154,21 @@ public FileTransferApiClientResponse delete(String requestId, String fileId) {
...
+     * the file-transfer-api into the provided outputStream.
+     * @param fileId The id used by the file-transfer-api to identify the file
+     * @param httpServletResponse The HttpServletResponse to stream the file to
+     * @return FileTransferApiClientResponse containing the http status
+     */
+    public FileTransferApiClientResponse download(String fileId, HttpServletResponse httpServletResponse) {
+        // TODO OBJ-200 replace this dummy response with the implementation.
+        FileTransferApiClientResponse fileTransferApiClientResponse = new FileTransferApiClientResponse();
+        fileTransferApiClientResponse.setFileId(fileId);
+        fileTransferApiClientResponse.setHttpStatus(HttpStatus.OK);
```
mwestacott, line 171: I know it's only a stub, but guessing this returns an empty file content (stream)? Could we maybe write some bytes (just some text) to the output so the file isn't empty if this ends up being called by the web client?

mattch1: Currently working on the controller and service that will call this - so I will merge this into what I am doing there - can I keep the stub as it is for now and then add the bytes in on that PR as I will be able to dev test it more fully?

mwestacott: Strictly speaking the work can/should be done in this PR since the implementation would be in this method that you've added - it's got nothing to do with the controller PR. You would need to copy an InputStream into the HTTP Servlet response a bit like in the Extensions FileTransferApiClient class and you could create the dummy input stream like this:

   `InputStream targetStream = new ByteArrayInputStream("DUMMY-ATTACHMENT".getBytes());`

And then test it's working in your unit-test.

mattch1: Alright added this to the stub.

PR #48, opened 2020-08-11
Feature/obj 197 endpoint controller

mattch1:
Download attachment endpoint linked to file transfer api download stub.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -56,6 +56,7 @@ public class ObjectionController {
     private static final String ERROR_500 = "Internal server error";
     private static final String COULD_NOT_DELETE = "Could not delete attachment";
     private static final String OBJECTION_NOT_PROCESSED = "Objection not processed";
+    private static final String DOWNLOAD_ERROR = "Download Error";
 
     private PluggableResponseEntityFactory responseEntityFactory;
     private IObjectionService objectionService;
@@ -464,7 +465,7 @@ public class ObjectionController {
         } catch (HttpClientErrorException | HttpServerErrorException e) {
             apiLogger.errorContext(
                     requestId,
-                   "",
+                    DOWNLOAD_ERROR,
                     e,
                     logMap
             );
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -35,7 +35,6 @@ import uk.gov.companieshouse.service.rest.response.PluggableResponseEntityFactor
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -460,7 +459,8 @@ public class ObjectionController {
                     "GET /{objectionId}/attachments/{attachmentId}/download request received",
                     logMap
             );
-            FileTransferApiClientResponse downloadServiceResult = objectionService.downloadAttachment(objectionId, attachmentId, response);
+            FileTransferApiClientResponse downloadServiceResult = objectionService.downloadAttachment(
+                    requestId, objectionId, attachmentId, response);
             return ResponseEntity.status(downloadServiceResult.getHttpStatus()).build();
         } catch (HttpClientErrorException | HttpServerErrorException e) {
             apiLogger.errorContext(
@@ -470,6 +470,21 @@ public class ObjectionController {
                     logMap
             );
             return ResponseEntity.status(e.getStatusCode()).build();
+        } catch (ServiceException e) {
+            apiLogger.errorContext(
+                    requestId,
+                    DOWNLOAD_ERROR,
+                    e,
+                    logMap
+            );
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+
+        } finally {
+            apiLogger.infoContext(
+                    requestId,
+                    "Finished DOWNLOAD /{objectionId}/attachments/{attachmentId}/download request",
+                    logMap
+            );
         }
     }
 }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -33,6 +33,7 @@ import uk.gov.companieshouse.service.rest.response.ChResponseBody;
 import uk.gov.companieshouse.service.rest.response.PluggableResponseEntityFactory;
 
 import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -437,6 +438,36 @@ public class ObjectionController {
                     "Finished DELETE /{objectionId}/attachments/{attachmentId} request",
                     logMap
             );
+        }
+    }
+
+    @GetMapping("/{objectionId}/attachments/{attachmentId}/download")
+    public ResponseEntity<Void> downloadAttachment(@PathVariable String companyNumber,
+                                                   @PathVariable String objectionId,
+                                                   @PathVariable String attachmentId,
+                                                   @RequestHeader(value = ERIC_REQUEST_ID_HEADER) String requestId,
+                                                   HttpServletResponse response) {
+        Map<String, Object> logMap = new HashMap<>();
+        logMap.put(LOG_COMPANY_NUMBER_KEY, companyNumber);
+        logMap.put(LOG_OBJECTION_ID_KEY, objectionId);
+
+        try{
+            apiLogger.infoContext(
+                    requestId,
+                    "GET /{objectionId}/attachments/{attachmentId}/download request received",
+                    logMap
+            );
+
+            objectionService.downloadAttachment(objectionId, attachmentId, response);
+            return new ResponseEntity<>(HttpStatus.OK);
+        } catch (HttpClientErrorException | HttpServerErrorException e) {
+            apiLogger.errorContext(
+                    requestId,
+                   "",
+                    e,
+                    logMap
+            );
+            return ResponseEntity.status(e.getStatusCode()).build();
         }
     }
 }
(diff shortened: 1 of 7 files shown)


ICoxCH:
Please check the sonar report for some unused imports, and logging an exception instead of e.printStackTrace()

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/file/FileTransferApiClient.java
```diff
@@ -167,14 +167,18 @@ public FileTransferApiClientResponse delete(String requestId, String fileId) {
...
         FileTransferApiClientResponse fileTransferApiClientResponse = new FileTransferApiClientResponse();
         fileTransferApiClientResponse.setFileId(fileId);
         fileTransferApiClientResponse.setHttpStatus(HttpStatus.OK);
 
         InputStream inputStream = new ByteArrayInputStream("This is a test".getBytes());
-        IOUtils.copy(inputStream, httpServletResponse.getOutputStream());
+        try {
+            IOUtils.copy(inputStream, httpServletResponse.getOutputStream());
+        } catch (IOException e) {
+            e.printStackTrace();
```
ICoxCH, line 180: Whats the thinking behind catching the exception here and not throwing it to the service for the service to deal with?

mattch1: No it should't really be caught here, it was only put there as part of the stop gap required to add some byte content to the stub so I thought little of leaving it there. However, there are lasting benefits to throwing an exception in the service and then throwing a service exception as we do with other methods  - this should be part of the stubbed structure so chnaged this.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -439,4 +442,49 @@ public ResponseEntity deleteAttachment(
...
+            return ResponseEntity.status(downloadServiceResult.getHttpStatus()).build();
+        } catch (HttpClientErrorException | HttpServerErrorException e) {
+            apiLogger.errorContext(
+                    requestId,
+                    DOWNLOAD_ERROR,
+                    e,
+                    logMap
+            );
+            return ResponseEntity.status(e.getStatusCode()).build();
+        } catch (ServiceException e) {
```
ICoxCH, line 473: Need a unit test to cover the service exception being thrown

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -286,6 +288,19 @@ private void deleteFromS3(String requestId, String attachmentId, Map<String, Obj
...
     }
 
+    public FileTransferApiClientResponse downloadAttachment(String requestId,
+                                                            String objectionId,
+                                                            String attachmentId,
+                                                            HttpServletResponse response) throws ServiceException {
+        Map<String, Object> logMap = buildLogMap(null, objectionId, attachmentId);
+        try {
+            return fileTransferApiClient.download(attachmentId, response);
+        } catch (IOException e) {
```
ICoxCH, line 298: Need a unit test for IOException being thrown

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
```diff
@@ -400,6 +406,85 @@ public void deleteAttachmentWhenUnableToDelete()
...
+                objectionController.downloadAttachment(
+                        COMPANY_NUMBER, OBJECTION_ID, ATTACHMENT_ID, REQUEST_ID, httpServletResponse);
+
+        assertEquals(HttpStatus.OK, responseEntity.getStatusCode());
+        assertNull(responseEntity.getBody());
+        assertTrue(responseEntity.getHeaders().isEmpty());
+    }
+
+    @Test
+    public void testReturnUnauthorizedStatusWhenInDownloadResponse() throws ServiceException {
```
ICoxCH, line 429: Could this test and the one below it have better names?
Perhaps testReturnUnauthorizedStatusWhenDownloadResponseUnauthorized

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
```diff
@@ -602,4 +605,21 @@ void deleteAttachmentHandleNullHttpStatusApiResponseOnDeleteAttachment() {
                 any());
     }
 
+    @Test
+    public void willCallFileTransferGatewayForDownload() throws IOException, ServiceException {
```
ICoxCH, line 609: Is it not calling the api? Not the gateway?

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
```diff
@@ -602,4 +605,21 @@ void deleteAttachmentHandleNullHttpStatusApiResponseOnDeleteAttachment() {
...
+
+        when(fileTransferApiClient.download(ATTACHMENT_ID, httpServletResponse)).thenReturn(dummyDownloadResponse);
+
+        FileTransferApiClientResponse downloadServiceResult = objectionService.downloadAttachment(
+                REQUEST_ID, OBJECTION_ID, ATTACHMENT_ID, httpServletResponse);
+
+        verify(fileTransferApiClient, only()).download(ATTACHMENT_ID, httpServletResponse);
+        verify(fileTransferApiClient, times(1)).download(ATTACHMENT_ID, httpServletResponse);
+
+        assertNotNull(downloadServiceResult);
```
ICoxCH, line 621: Can we check that the status of the result is the same as the dummyDownloadResponse? Just to confirm the ObjectionService doesn't modify it

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
mattch1: I don't really want to repeat the word "Unauthroized" - the name's not that bad surely?

ICoxCH: To me it wasn't immediately clear what was supposed to be "in" the download response.
My initial thoughts were it was short for invalid response.
Maybe get another opinion on this one, but it's not overly important

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: We shold bear in mind that this is really only part of the stop gap and is likely to change.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
mwestacott: 'testReturnUnauthorizedStatusWhenDownloadFailsAuthorizationChecks' perhaps? (and 'testReturnForbiddenStatusWhenNotAllowedToDownloadAnInfectedFile' maybe)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -439,4 +442,49 @@ public ResponseEntity deleteAttachment(
...
+            apiLogger.infoContext(
+                    requestId,
+                    "GET /{objectionId}/attachments/{attachmentId}/download request received",
+                    logMap
+            );
+            FileTransferApiClientResponse downloadServiceResult = objectionService.downloadAttachment(
+                    requestId, objectionId, attachmentId, response);
+            return ResponseEntity.status(downloadServiceResult.getHttpStatus()).build();
+        } catch (HttpClientErrorException | HttpServerErrorException e) {
+            apiLogger.errorContext(
```
markpit, line 466: In PTF I was getting issues with re-using the logMap as the logger would modify the map and place the message into it. On subsequent calls it would see the map already had a message from the first call and not replace it, resulting in the old message being displayed. If this is logging correctly I don't understand how it is, I think we should use a new map for each log call. You can see how it works here https://github.com/companieshouse/structured-logging/blob/master/src/main/java/uk/gov/companieshouse/logging/StructuredLogger.java

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
markpit: can't we just call it testReturnUnauthorizedStatusForDownload

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: This looks like something we need to be changing more widely there is a ticket to do this on the backlog OBJ-141 - i'll add this to the ticketas it provides some guidence as to what we need to be aiming for.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
markpit: I think this is ok, service calls fileTransferApiClient which returns it's own FileTransferApiClientResponse object

markpit: ah, i think you meant the test title. Yes, we don't have a gateway we have the FileTransferApiClient so would be better to rename the title.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: Added a test.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: Added a test

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
markpit: That story is to centralise the logging. I'll create a new one to ensure it does not get ignored.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
mattch1: Shorted the name to something clearer

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
mattch1: Changed name.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
mattch1: Now checking the status

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: Ok we should link it to the 141 issue thought

PR #53, opened 2020-08-20
OBJ-7 Add user authentication interceptor

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/authorization/UserAuthorizationInterceptor.java
```diff
@@ -0,0 +1,65 @@
...
+        apiLogger.debugContext(requestId, "Checking current user is authorised for objection id");
+        final String user = request.getHeader(ERICHeaderFields.ERIC_AUTHORISED_USER);
+        final String objectionId = getObjectionId(request.getRequestURI());
+
+        final Objection objection = objectionService.getObjection(requestId, objectionId);
+
+        final String createdByUserEmail = objection.getCreatedBy().getEmail();
+        final String requestUserEmail = ericHeaderParser.getEmailAddress(user);
+
+        if(createdByUserEmail.equals(requestUserEmail)) {
```
mwestacott, line 41: Not sure this will allow a dissolution team member to download an attachment related to an objection that was created by someone else.

mwestacott:
Also wondering if you're intending to address the 'company matches objection' comment/observation on OBJ-7 in this PR or another one?

ICoxCH:
> Also wondering if you're intending to address the 'company matches objection' comment/observation on OBJ-7 in this PR or another one?

It'll be in this pr, maybe should of used a draft pr but wanted some early eyes on so we could decide what direction to take with regards to the multiple database calls this could bring up

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/authorization/UserAuthorizationInterceptor.java
ICoxCH: I think you're right, i'll make sure thats addressed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/authorization/UserAuthorizationInterceptor.java
```diff
@@ -0,0 +1,65 @@
...
+            apiLogger.debugContext(requestId, "User is authorised for objection id");
+            return true;
+        }
+
+        apiLogger.debugContext(requestId, "User not authorised for objection id");
+        response.setStatus(HttpStatus.UNAUTHORIZED.value());
+        return false;
+    }
+
+    private String getObjectionId(String requestURI) throws ServiceException{
```
mwestacott, line 51: I'm a bit uneasy about this logic (living here, I suppose). The URIs are all specified in the controller and I suspect they could change there and the unit test for this class would still pass. Perhaps an integration test would mitigate the risk...

Alternatively, the controller methods benefit from the @PathVariable annotation which automatically extracts the objection id. Can we somehow make use of that approach in this interceptor too?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/authorization/UserAuthorizationInterceptor.java
markpit: also thinking we should allow a support team person to re-run failures via patch. Dn't now if there's a requirement for it yet though.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/InterceptorConfig.java
```diff
@@ -6,23 +6,64 @@
...
     public void addInterceptors(InterceptorRegistry registry) {
         registry.addInterceptor(attachmentDownloadAuthorizationInterceptor(logger))
-            .addPathPatterns("/**/attachments/**/download");
+            .addPathPatterns(ATTACHMENTS_DOWNLOAD_PATH);
+        registry.addInterceptor(objectionInterceptor(objectionService, logger))
+                .addPathPatterns(STRIKE_OFF_OBJECTIONS_OBJECTION_ID)
+                .excludePathPatterns(ATTACHMENTS_DOWNLOAD_PATH);
+        registry.addInterceptor(companyNumberInterceptor(logger))
+                .addPathPatterns(STRIKE_OFF_OBJECTIONS_OBJECTION_ID)
+                .excludePathPatterns(ATTACHMENTS_DOWNLOAD_PATH);
```
mwestacott, line 64: Wondering why we don't do the company number - objection id check on this path too? Both ids are present when trying to download an attachment and should/can be checked, no?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/CompanyNumberInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public CompanyNumberInterceptor(ApiLogger apiLogger) {
+        this.apiLogger = apiLogger;
+    }
+
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws Exception {
+        final String requestId = request.getHeader(ERICHeaderFields.ERIC_REQUEST_ID);
+        apiLogger.debugContext(requestId, "Checking provided company number matches objection company number");
+        final Objection objection = (Objection) request.getAttribute("objection");
```
mwestacott, line 26: Would recommend defining and using a constant for the 'objection' attribute name. Used here and in the other interceptor class (possibly other places?).

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/CompanyNumberInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+
+        Map<String, String> pathVariables =
+                (Map<String, String>) request.getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+
+        final String companyNumber = pathVariables.get("companyNumber");
```
mwestacott, line 31: A constant here for the path variable key name too. Used here and in controller (maybe elsewhere?).

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/CompanyNumberInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+
+        boolean result = false;
```
mwestacott, line 33: Would prefer a more descriptive variable name here. How about 'companyNumberMatches'?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/authorization/UserAuthorizationInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+    public UserAuthorizationInterceptor(ApiLogger apiLogger, ERICHeaderParser ericHeaderParser) {
+        this.apiLogger = apiLogger;
+        this.ericHeaderParser = ericHeaderParser;
+    }
+
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws Exception {
+
+        final String requestId = request.getHeader(ERICHeaderFields.ERIC_REQUEST_ID);
+        apiLogger.debugContext(requestId, "Checking current user is authorised for objection id");
```
mwestacott, line 27: Minor, but user is authorised to do something with the objection and not the 'objection id'.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/authorization/AuthorizationIntegrationTest.java
```diff
@@ -1,4 +1,4 @@
-package uk.gov.companieshouse.api.strikeoffobjections.authorization;
+package uk.gov.companieshouse.api.strikeoffobjections.interceptor.authorization;
```
mwestacott, line 1: This integration test was specifically targeting the original interceptor I wrote. I think extending it my adding more tests to check all the new interceptors work in the context of the Spring Boot app - and together - would be very useful...

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/authorization/UserAuthorizationInterceptor.java
markpit: i think we need the id in the message to identify what objection they were trying to access. maybe stick it on the end? 'Checking if current user is authorised to access this objection that has an id of '

mwestacott: Sounds good.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/CompanyNumberInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+        final String requestId = request.getHeader(ERICHeaderFields.ERIC_REQUEST_ID);
+        apiLogger.debugContext(requestId, "Checking provided company number matches objection company number");
+        final Objection objection = (Objection) request.getAttribute(InterceptorConstants.OBJECTION_ATTRIBUTE);
+
+        Map<String, String> pathVariables =
+                (Map<String, String>) request.getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+
+        final String companyNumber = pathVariables.get(InterceptorConstants.COMPANY_NUMBER_PATH_VARIABLE);
+
+        boolean companyNumberMatches = false;
```
markpit, line 33: line 33 - 36 you could make a private function eg. doCompanyNumbersMatch(companyNumber, objection)
then line 38 can be if (!doCompanyNumbersMatch(companyNumber, objection) {

markpit (approved):

mwestacott (approved):

PR #54, opened 2020-08-24
OBJ-227 Ineligible http code

mattch1:
When create is called and a company is not eligible an HTTP 406 code: NOT_ACEPTABLE is returned. One of 4 status is possible 3 corresponding to the ineligibility codes on OBJ-208 and 1 if no codes are present as described on OBJ-209.

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -409,25 +409,20 @@
         }
       }
     },
+    "Ineligible": {
+      "type": "object",
+      "properties": {
+        "status": {
+          "$ref": "#/definitions/Status"
+        }
+      }
+    },
     "Status": {
       "type": "string",
       "enum": [
         "OPEN",
         "SUBMITTED",
-        "PROCESSED"
-      ]
-    },
-    "Ineligible": {
-      "type": "object",
-      "properties": {
-        "status": {
-          "$ref": "#/definitions/IneligibilityStatus"
-        }
-      }
-    },
-    "IneligibilityStatus": {
-      "type": "string",
-      "enum": [
+        "PROCESSED",
         "INELIGIBLE_COMPANY_STRUCK_OFF",
         "INELIGIBLE_NO_DISSOLUTION_ACTION"
       ]
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -429,9 +429,7 @@
       "type": "string",
       "enum": [
         "INELIGIBLE_COMPANY_STRUCK_OFF",
-        "INELIGIBLE_COMPANY_SECOND_GAZETTE",
-        "INELIGIBLE_COMPANY_CONVERTED",
-        "INELIGIBLE_COMPANY_NO_ELIGIBILITY_CODE"
+        "INELIGIBLE_NO_DISSOLUTION_ACTION"
       ]
     }
   }
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -54,7 +54,7 @@
               "$ref": "#/definitions/ResourceCreated"
             }
           },
-          "406": {
+          "400": {
             "description": "Company was ineligible for objecting to strike off",
             "schema": {
               "$ref": "#/definitions/Ineligible"
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -55,7 +55,7 @@
             }
           },
           "400": {
-            "description": "Company was ineligible for objecting to strike off",
+            "description": "Company is ineligible for objecting to strike off",
             "schema": {
               "$ref": "#/definitions/Ineligible"
             }
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -426,7 +426,9 @@
       "enum": [
         "OPEN",
         "SUBMITTED",
-        "PROCESSED"
+        "PROCESSED",
+        "INELIGIBLE_COMPANY_STRUCK_OFF",
+        "INELIGIBLE_NO_DISSOLUTION_ACTION"
       ]
     }
   }
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -53,6 +53,12 @@
             "schema": {
               "$ref": "#/definitions/ResourceCreated"
             }
+          },
+          "406": {
+            "description": "Company was ineligible for objecting to strike off",
+            "schema": {
+              "$ref": "#/definitions/Ineligible"
+            }
           }
         }
       }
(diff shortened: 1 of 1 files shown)


spec/schema.json
```diff
@@ -53,6 +53,12 @@
             "schema": {
               "$ref": "#/definitions/ResourceCreated"
             }
+          },
+          "406": {
```
ICoxCH, line 57: According to Mozilla: 406 Not Acceptable client error response code indicates that the server cannot produce a response matching the list of acceptable values defined in the request's proactive content negotiation headers, and that the server is unwilling to supply a default representation.

As this error doesn't relate to content types i'd be hesitant about using this code. I think a simple 400, with a body containing an ineligible code would be more than descriptive enough

mattch1: There don't seem to be any other 4xx errors that are suitable so will opt for 400.

ICoxCH (approved):

spec/schema.json
```diff
@@ -53,6 +53,12 @@
             "schema": {
               "$ref": "#/definitions/ResourceCreated"
             }
+          },
+          "400": {
+            "description": "Company was ineligible for objecting to strike off",
```
mwestacott, line 58: Minor, but suggest changing the 'was' to 'is' in the description. It's at the point the end-point is called that the company is ineligible - we're not doing an historical check.

mattch1: Changed this to "is"

spec/schema.json
```diff
@@ -410,6 +416,23 @@
...
+      "properties": {
+        "status": {
+          "$ref": "#/definitions/IneligibilityStatus"
+        }
+      }
+    },
+    "IneligibilityStatus": {
+      "type": "string",
+      "enum": [
+        "INELIGIBLE_COMPANY_STRUCK_OFF",
```
mwestacott, line 431: I thought we agreed in the meeting that what we'd return was a code to the web to allow for the screen navigation (not the finer-grained action codes in the story) - so we'd only need two codes in theory, INELIGIBLE_COMPANY_STRUCK_OFF for OBJ-208 and (probably) INELIGIBLE_NO_DISSOLUTION_ACTION for OBJ-209 if we're making the spec change for that story too now.

mattch1: It's not something we specified on the ticket so I think these were just ideas we were playing with. The logic would be much the same in the web because we would look for "INELIGIBLE_COMPANY_NO_ELIGIBILITY_CODE" in which case we redirect to the OBJ-209 screen and if not then we go to the OBJ-208 screen. Having these codes would allow us to log or provide a reason on the 208 error page should that become a requirement.

spec/schema.json
```diff
@@ -410,6 +416,23 @@
...
+        "INELIGIBLE_COMPANY_SECOND_GAZETTE",
+        "INELIGIBLE_COMPANY_CONVERTED",
```
markpit, line 433: i think INELIGIBLE_COMPANY_CONVERTED  should be INELIGIBLE_COMPANY_CONVERTED_OR_CLOSED

mattch1: Yeh, I didn't want to make them too long so tried to abbreviate as best I could, but can change if we are happy with having the 4 codes - see Martin's comment above.

spec/schema.json
mwestacott: Worth checking with @markpit as he wrote up the sub-task. I thought we only bothered adding the ineligible code to the 400 response in order to distinguish between the 208 and 209 stories, in which case two codes would be sufficient. I guess it's where we want that business logic around the action codes to live - in the web or the API... (the finer-grained action code should definitely be logged in the API though)

markpit: I thought it would be whether the action code was in one of the groups specified in the story. The api would get the action code, check what group it was in and return the status enum value for that ineligible group. That status would also be applied to the database record and the action code stored on it as well.
   If we don't group them, if more codes are added/subtracted in future we would need to update the spec and create new enum value for each ineligible action code and modify the web app to look for that new status. I think if we go with the groups approach it will give us more flexibility if they do want to add/subtract codes to a group and at this point the web is on the only user of the api (which needs 2 codes to show 2 ineligible screens)

mattch1: Any thoughts on what these groups should be? Here's a suggestion:

ELIGIBLE_COMPANY
INELIGIBLE_COMPANY
NO_ACTION_FOUND

markpit: i'd go with what Martin said and have 2 statuses to indicate the 2 groups of ineligible codes, perhaps go with martin's suggestions above for the statuses (or similar)

markpit: if it's eligible, we would return a 200 ok http code and continue as normal (so no need to return a status if 200 ok)

mattch1: ... and these are the codes we should add to the Objection model for OBJ-230?

mattch1: Gone with Martin's suggestion and grouped them.

markpit: We'd need them added to the ObjectionStatusEnum which you can do as part of OBJ-230 and we need a new field on Objection called actionCode to store the actual action code against the objection. We's also need to update the builders for actionCode as well.

mattch1: So we store both the CHIPS code and our group codes on the object?

markpit: yes we'll store both although the objection status field already exists, we'll just populate it with the new values. And add a new action code field.

markpit (approved):

spec/schema.json
```diff
@@ -410,6 +416,21 @@
...
+    },
+    "Ineligible": {
+      "type": "object",
+      "properties": {
+        "status": {
+          "$ref": "#/definitions/IneligibilityStatus"
+        }
+      }
+    },
+    "IneligibilityStatus": {
```
markpit, line 428: I approved but spotted something, the INELIGIBLE_XXXX status should be added to 
"Status": {
      "type": "string",
      "enum": [
        "OPEN",
        "SUBMITTED",
        "PROCESSED"
      ]
    },

so we don't need a new class of statuses

mwestacott (approved):

spec/schema.json
mattch1: Removed new class added to old class.

mwestacott (changes requested):
This looks wrong. I think you just needed to add those two new codes to 'Status' as now it reads as if the ineligibility check could also return OPEN, SUBMITTED or PROCESSED (which it can't). IneligibilityStatus is still required.

mattch1:
That should not be a problem we would not ever set the error to return OPEN, SUBMITTED or PROCESSED.

mattch1:
Updated to an inline enum - i don’t think we can get away with not mentioning enum entirely.

spec/schema.json
```diff
@@ -403,6 +409,18 @@
...
+          "enum": [
+            "INELIGIBLE_COMPANY_STRUCK_OFF",
+            "INELIGIBLE_NO_DISSOLUTION_ACTION"
+          ]
+        }
+      }
+    },
     "Status": {
       "type": "string",
       "enum": [
```
mwestacott, line 426: We still need those two new statuses in the Status type, as per Mark's original comment.

spec/schema.json
```diff
@@ -403,6 +409,18 @@
         }
       }
     },
+    "Ineligible": {
+      "type": "object",
+      "properties": {
+        "status": {
+          "type": "string",
+          "enum": [
```
mwestacott, line 417: 👍

spec/schema.json
mattch1: Reintroduced these.

mwestacott (approved):

PR #55, opened 2020-08-24
OBJ-230 action codes for objections

mattch1:
Updated status code and added an action codes field and enum to the Objection model.


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
```diff
@@ -78,6 +85,8 @@ public Objection build() {
     private List<Attachment> attachments = new ArrayList<>();
     @Field("status")
     private ObjectionStatus status;
+    @Field("action_code")
+    private ActionCode actionCode;
```
markpit, line 89: we want to store the actual number in mongo

mattch1: Changed field type.

ICoxCH (approved):
Looks good to me

mwestacott (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ActionCode.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.companieshouse.api.strikeoffobjections.model.entity;
+
+public enum ActionCode {
```
markpit, line 3: we don't need the enum anymore if we are putting the number into the DB

mattch1: Removed this.

markpit (approved):

PR #56, opened 2020-08-25
OBJ-232 api controller return obj status for eligibility

mattch1:
Updated service call to return a DTO rather than a string which contains a status as well as an id.

The controller will return an hhtp 400 error and a body ctaining a dto with the objection status in it.
For the response have used a normal body rather than an error body as this takes an error container object and will mean alterations to the agreed schema.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -146,7 +146,8 @@ public class ObjectionController {
         );
 
         try {
-            ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            Objection objection = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionResponseDTO response = objectionMapper.objectionEntityToObjectionResponseDTO(objection);
             ObjectionStatus objectionStatus = response.getStatus();
             if (objectionStatus.isIneligible()) {
                return new ResponseEntity<>(ChResponseBody.createNormalBody(response), HttpStatus.BAD_REQUEST);
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -147,12 +147,13 @@ public class ObjectionController {
 
         try {
             Objection objection = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionResponseDTO response = objectionMapper.objectionEntityToObjectionResponseDTO(objection);
-            ObjectionStatus objectionStatus = response.getStatus();
+            ObjectionStatus objectionStatus = objection.getStatus();
             if (objectionStatus.isIneligible()) {
+               ObjectionResponseDTO response = new  ObjectionResponseDTO();
+               response.setStatus(objectionStatus);
                return new ResponseEntity<>(ChResponseBody.createNormalBody(response), HttpStatus.BAD_REQUEST);
             }
-            return responseEntityFactory.createResponse(ServiceResult.created(response));
+            return responseEntityFactory.createResponse(ServiceResult.created(new ObjectionResponseDTO(objection.getId())));
         } catch (Exception e) {
             apiLogger.errorContext(
                     requestId,
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -147,8 +147,8 @@ public class ObjectionController {
 
         try {
             ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionStatus obStat = response.getStatus();
-            if (obStat != null && obStat.isEligibilityError()) {
+            ObjectionStatus objectionStatus = response.getStatus();
+            if (objectionStatus != null && objectionStatus.isIneligibleStatus()) {
                return new ResponseEntity<>(ChResponseBody.createNormalBody(response), HttpStatus.BAD_REQUEST);
             }
             return responseEntityFactory.createResponse(ServiceResult.created(response));
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -23,6 +23,7 @@ import uk.gov.companieshouse.api.strikeoffobjections.exception.ObjectionNotFound
 import uk.gov.companieshouse.api.strikeoffobjections.file.FileTransferApiClientResponse;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Attachment;
 import uk.gov.companieshouse.api.strikeoffobjections.model.entity.Objection;
+import uk.gov.companieshouse.api.strikeoffobjections.model.entity.ObjectionStatus;
 import uk.gov.companieshouse.api.strikeoffobjections.model.patch.ObjectionPatch;
 import uk.gov.companieshouse.api.strikeoffobjections.model.response.AttachmentResponseDTO;
 import uk.gov.companieshouse.api.strikeoffobjections.model.response.ObjectionResponseDTO;
@@ -145,8 +146,11 @@ public class ObjectionController {
         );
 
         try {
-            String objectionId = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionResponseDTO response = new ObjectionResponseDTO(objectionId);
+            ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionStatus obStat = response.getStatus();
+            if (obStat != null && obStat.isEligibilityError()) {
+               return new ResponseEntity<>(ChResponseBody.createNormalBody(response), HttpStatus.BAD_REQUEST);
+            }
             return responseEntityFactory.createResponse(ServiceResult.created(response));
         } catch (Exception e) {
             apiLogger.errorContext(
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -149,7 +149,7 @@ public class ObjectionController {
             Objection objection = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
             ObjectionStatus objectionStatus = objection.getStatus();
             if (objectionStatus.isIneligible()) {
-               ObjectionResponseDTO response = new  ObjectionResponseDTO();
+               ObjectionResponseDTO response = new ObjectionResponseDTO();
                response.setStatus(objectionStatus);
                return new ResponseEntity<>(ChResponseBody.createNormalBody(response), HttpStatus.BAD_REQUEST);
             }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -148,7 +148,7 @@ public class ObjectionController {
         try {
             ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
             ObjectionStatus objectionStatus = response.getStatus();
-            if (objectionStatus != null && objectionStatus.isIneligibleStatus()) {
+            if (objectionStatus.isIneligibleStatus()) {
                return new ResponseEntity<>(ChResponseBody.createNormalBody(response), HttpStatus.BAD_REQUEST);
             }
             return responseEntityFactory.createResponse(ServiceResult.created(response));
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -148,7 +148,7 @@ public class ObjectionController {
         try {
             ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
             ObjectionStatus objectionStatus = response.getStatus();
-            if (objectionStatus.isIneligibleStatus()) {
+            if (objectionStatus.isIneligible()) {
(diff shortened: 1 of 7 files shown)


ICoxCH (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -145,8 +146,11 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
         );
 
         try {
-            String objectionId = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionResponseDTO response = new ObjectionResponseDTO(objectionId);
+            ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionStatus obStat = response.getStatus();
```
markpit, line 150: can you rename obStat to objectionStatus to improve readability

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
```diff
@@ -5,5 +5,10 @@
     PROCESSED,
     SUBMITTED,
     INELIGIBLE_COMPANY_STRUCK_OFF,
-    INELIGIBLE_NO_DISSOLUTION_ACTION
+    INELIGIBLE_NO_DISSOLUTION_ACTION;
+
+    public boolean isEligibilityError() {
```
markpit, line 10: I wouldn't class it as an error as they can be ineligible without causing an error. How about isIneligibleStatus()?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
```diff
@@ -5,5 +5,10 @@
...
+        return this == INELIGIBLE_NO_DISSOLUTION_ACTION ||
```
markpit, line 11: I would have a static Set of the ineligible statuses and check if that Set contains it to prevent the if statement growing in future.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -72,9 +73,13 @@ public ObjectionService(ObjectionRepository objectionRepository,
...
 
     @Override
-    public String createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails) {
+    public ObjectionResponseDTO createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails) {
         Map<String, Object> logMap = buildLogMap(companyNumber, null, null);
         logger.infoContext(requestId, "Creating objection", logMap);
+        ObjectionResponseDTO response = new ObjectionResponseDTO();
+
+        // TODO OBJ-231 process query result and return eligibility status
+        ObjectionStatus obStat = ObjectionStatus.OPEN;
```
markpit, line 82: can you rename obStat to objectionStatus

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
markpit, line 11: This new function needs a unit test, need to add an ObjectionStatusTest class

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
```diff
@@ -5,5 +5,10 @@
     PROCESSED,
     SUBMITTED,
     INELIGIBLE_COMPANY_STRUCK_OFF,
-    INELIGIBLE_NO_DISSOLUTION_ACTION
+    INELIGIBLE_NO_DISSOLUTION_ACTION;
```
markpit, line 8: can you keep these in alphabetical order

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -145,8 +146,11 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
...
+            if (obStat != null && obStat.isEligibilityError()) {
```
mwestacott, line 151: Can this really ever be null at this point in a legitimate scenario? Not sure it can be and if that's the case and it is somehow null, this indicates that something unexpected has happened - we would probably want the code to fail 'noisily', so blowing up with a NullPointerException would be preferable and perfectly fine. Basically, what I'm saying is, I don't think that the check for 'obStat != null' is required.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
mwestacott: Or maybe just 'isIneligible()' (status is already present in the class name...)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
mattch1: Reordered

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
mattch1: Three new unit tests

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: Renamed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
mattch1: Renamed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: Renamed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionStatus.java
```diff
@@ -1,9 +1,15 @@
...
+    INELIGIBLE_COMPANY_STRUCK_OFF,
+    INELIGIBLE_NO_DISSOLUTION_ACTION,
     OPEN,
     PROCESSED,
-    SUBMITTED,
-    INELIGIBLE_COMPANY_STRUCK_OFF,
-    INELIGIBLE_NO_DISSOLUTION_ACTION
+    SUBMITTED;
+
+    public boolean isIneligibleStatus() {
```
mwestacott, line 11: How about just 'isIneligible()' for this method name? Status is already present in/clear from the class name.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -145,8 +146,11 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
         );
 
         try {
-            String objectionId = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionResponseDTO response = new ObjectionResponseDTO(objectionId);
+            ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionStatus objectionStatus = response.getStatus();
+            if (objectionStatus != null && objectionStatus.isIneligibleStatus()) {
```
mwestacott, line 151: Can this really ever be null at this point in a legitimate scenario? Not sure it can be and if that's the case and it is somehow null, this indicates that something unexpected has happened - we would probably want the code to fail 'noisily', so blowing up with a NullPointerException would be preferable and perfectly fine. Basically, what I'm saying is, I don't think that the check for 'obStat != null' is required.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -145,8 +146,11 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
         );
 
         try {
-            String objectionId = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionResponseDTO response = new ObjectionResponseDTO(objectionId);
+            ObjectionResponseDTO response = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionStatus objectionStatus = response.getStatus();
+            if (objectionStatus.isIneligibleStatus()) {
```
mwestacott, line 151: @mattch1 did you see my other comment about renaming this method to just 'isIneligible()'? Would read better here.

mattch1: Seen an updated it.

mwestacott (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mwestacott: Thanks.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IObjectionService.java
```diff
@@ -10,13 +10,14 @@
...
 import uk.gov.companieshouse.api.strikeoffobjections.model.patch.ObjectionPatch;
+import uk.gov.companieshouse.api.strikeoffobjections.model.response.ObjectionResponseDTO;
 import uk.gov.companieshouse.service.ServiceException;
 import uk.gov.companieshouse.service.ServiceResult;
 
 import javax.servlet.http.HttpServletResponse;
 
 public interface IObjectionService {
-    String createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails);
+    ObjectionResponseDTO createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails);
```
markpit, line 20: I added a comment yesterday but it has disappeared, I don't think we should be returning the DTO as it is intended for converting to json for the api response. I'd prefer a new wrapper class that belongs in the same package as the service to wrap the results from createObjection.

mattch1: Not sure I follow - won't we just end up with more classes than we need? I'm not sure why we can't just use the DTO.

mwestacott: Could we just return/use the 'Objection' entity class instead then (and map it). Similar to how the '_getObjection()_' method works?

mattch1: Then we just have more code than we need -do we really want extra mapping?

mattch1: The mapping is probably the best solution will look into this.

markpit: We need to think of these components as being independent, by returning the DTO used in the controller layer, we are tying the service to the controller layer. If another team want to make changes to this code base in 2 years time, they will look at the createObjection return type and wonder why it is full of fields that aren't used and full of json annotations. Plus, more, smaller classes is better than fewer bigger classes, if each one has it's own purpose (separation of concerns I think it is called)

markpit: We could just return Objection as martin said, it would be logical for a method called createObjection to return an objection.

mwestacott: The mapper is already written so shouldn't be much more work.

mattch1: Service returns the objection and then in the controller this is mapped to the dto and returned.

mwestacott (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -145,8 +146,12 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
         );
 
         try {
-            String objectionId = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionResponseDTO response = new ObjectionResponseDTO(objectionId);
+            Objection objection = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionResponseDTO response = objectionMapper.objectionEntityToObjectionResponseDTO(objection);
```
markpit, line 150: we don't need to call the mapper, this will break the api spec as it looks like it will now return a full objection object in json in the http response body. The objection mapper will map all the fields. For successful, we should just return the id (see previous version of this method) and if ineligible we want just the status. Can't we just add status into the ObjectionResponseDTO constructor and populate it with the status from the returned Objection?

mattch1: In all the excitment of making the other changes forgot all about the spec - reworked this.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -145,9 +146,14 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
         );
 
         try {
-            String objectionId = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
-            ObjectionResponseDTO response = new ObjectionResponseDTO(objectionId);
-            return responseEntityFactory.createResponse(ServiceResult.created(response));
+            Objection objection = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionStatus objectionStatus = objection.getStatus();
+            if (objectionStatus.isIneligible()) {
+               ObjectionResponseDTO response = new  ObjectionResponseDTO();
```
mwestacott, line 152: Double space before 'ObjectionResponseDTO()'

mwestacott (approved):
Spotted something minor, looking good apart from that. Good spot @markpit on that response JSON.

markpit (commented):
Looks good, just a couple of new code smells in Sonar with the new changes: http://code-analysis.platform.aws.chdev.org:9000/dashboard?id=uk.gov.companieshouse%3Astrike-off-objections-api&pullRequest=56

markpit (approved):

PR #57, opened 2020-08-25
OBJ-231 validate action code

markpit:
Validation rules for OBJ-231
Adding the ActionCodeValidator and validation rules.
Reads the action codes to validate against from config.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
@@ -17,6 +17,8 @@ import java.util.List;
 @Configuration
 public class ValidationConfig {
 
+    private static final String CONFIG_MESSAGE_PREFIX = "CHS ENV CONFIG -";
+
     @Value("#{'${ACTION_CODES_COMPANY_STRUCK_OFF}'.split(',')}")
     private List<Long> companyStruckOffActionCodes;
 
@@ -42,10 +44,9 @@ public class ValidationConfig {
 
     @PostConstruct
     private void logActionCodeValidatorConfigValues() {
-        String messagePrefix = "CHS ENV CONFIG -";
         apiLogger.info(
-                String.format("%s ACTION_CODES_COMPANY_STRUCK_OFF = %s", messagePrefix, companyStruckOffActionCodes));
+                String.format("%s ACTION_CODES_COMPANY_STRUCK_OFF = %s", CONFIG_MESSAGE_PREFIX, companyStruckOffActionCodes));
         apiLogger.info(
-                String.format("%s ACTION_CODES_STRIKE_OFF_NOTICE = %s", messagePrefix, strikeOffNoticeActionCodes));
+                String.format("%s ACTION_CODES_STRIKE_OFF_NOTICE = %s", CONFIG_MESSAGE_PREFIX, strikeOffNoticeActionCodes));
     }
 }
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
@@ -1,5 +1,6 @@
 package uk.gov.companieshouse.api.strikeoffobjections.config;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -9,6 +10,7 @@ import uk.gov.companieshouse.api.strikeoffobjections.validation.AllowedValuesVal
 import uk.gov.companieshouse.api.strikeoffobjections.validation.DisallowedValuesValidationRule;
 import uk.gov.companieshouse.api.strikeoffobjections.validation.ValidationRule;
 
+import javax.annotation.PostConstruct;
 import java.util.Arrays;
 import java.util.List;
 
@@ -21,10 +23,11 @@ public class ValidationConfig {
     @Value("#{'${ACTION_CODES_STRIKE_OFF_NOTICE}'.split(',')}")
     private List<Long> strikeOffNoticeActionCodes;
 
-    @Bean
-    public List<ValidationRule<Long>> getActionCodeValidationRules(ApiLogger apiLogger) {
-        logActionCodeValidatorConfigValues(apiLogger);
+    @Autowired
+    private ApiLogger apiLogger;
 
+    @Bean
+    public List<ValidationRule<Long>> getActionCodeValidationRules() {
         return Arrays.asList(
                 new DisallowedValuesValidationRule<>(
                         companyStruckOffActionCodes,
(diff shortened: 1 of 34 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
```diff
@@ -0,0 +1,37 @@
...
+import uk.gov.companieshouse.api.strikeoffobjections.validation.DisallowedValuesValidationRule;
+import uk.gov.companieshouse.api.strikeoffobjections.validation.ValidationRule;
+
+import java.util.Arrays;
+import java.util.List;
+
+@Configuration
+public class ValidationConfig {
+
+    @Value("#{'${ACTION_CODES_COMPANY_STRUCK_OFF}'.split(',')}")
```
mwestacott, line 18: Nice. Didn't know you could inline the split() call.

markpit: good ole google, it even converts them into Longs

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
```diff
@@ -0,0 +1,37 @@
...
+    private List<Long> strikeOffNoticeActionCodes;
+
+    @Bean
+    public List<ValidationRule<Long>> getActionCodeValidationRules(ApiLogger apiLogger) {
+        return Arrays.asList(
+                new DisallowedValuesValidationRule<>(
+                        companyStruckOffActionCodes,
+                        ObjectionStatus.INELIGIBLE_COMPANY_STRUCK_OFF,
+                        apiLogger),
+                new AllowedValuesValidationRule<>(
```
mwestacott, line 31: This rule (or the name of it) seems odd. Why would the 'no current dissolution' action codes be assigned to an 'allowed' rule?

markpit: The rule will check if the input value is in a list of allowable values. So if the action code is not in the allowed list, it will throw a validation exception with that status. The requirement was to check if the code was not one of the list.

markpit: maybe i could rename them to DisallowedValue (no s) and same for the other one

mwestacott: Understand this now...

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
```diff
@@ -0,0 +1,37 @@
...
+public class ValidationConfig {
+
+    @Value("#{'${ACTION_CODES_COMPANY_STRUCK_OFF}'.split(',')}")
+    private List<Long> companyStruckOffActionCodes;
+
+    @Value("#{'${ACTION_CODES_STRIKE_OFF_NOTICE}'.split(',')}")
+    private List<Long> strikeOffNoticeActionCodes;
+
+    @Bean
+    public List<ValidationRule<Long>> getActionCodeValidationRules(ApiLogger apiLogger) {
```
mwestacott, line 25: I think we need to log the action codes that are read from config and being assigned to the various rules at some point - ideally only once and not every time that rules are run. Maybe that's in this method somewhere?

markpit: i've got the rules logging on a failure but only set to debug. I could up it to info. I'll also look at logging the sets of numbers only once, probably happen on startup when the spring beans get autowired.

markpit: I've added a private method to log out the values that is called when the getActionCodeValidationRules Bean is initialised on startup. I looked at having a constructor in this config class but the ApiLogger bean wasn't available to be injected into the constructor at that point during initialisation.

markpit: just occurred to me I might be able to annotate a method with PostConstruct that will then have access to the the autowired beans (ApiLogger). I'll give that at try.

markpit: Got it working with a @PostConstruct method in the config class. PR updated.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
```diff
@@ -0,0 +1,51 @@
...
+                new AllowedValuesValidationRule<>(
+                        strikeOffNoticeActionCodes,
+                        ObjectionStatus.INELIGIBLE_NO_DISSOLUTION_ACTION,
+                        apiLogger)
+        );
+    }
+
+    @PostConstruct
+    private void logActionCodeValidatorConfigValues() {
+        String messagePrefix = "CHS ENV CONFIG -";
```
mwestacott, line 45: Could have this message prefix as a class constant but there again this method is (in theory) only ever called once.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/validation/AllowedValuesValidationRule.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public void validate(T input, String logContext) throws ValidationException {
+        if (!allowableValues.contains(input)) {
+            apiLogger.debugContext(logContext, String.format("%s is not an allowed value", input));
+            throw new ValidationException(failureStatus);
+        }
+    }
+
```
mwestacott, line 29: Blank line here can go.

mwestacott (approved):
One thought and one very minor spot but all good.

mwestacott (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
```diff
@@ -0,0 +1,52 @@
...
+    private List<Long> companyStruckOffActionCodes;
+
+    @Value("#{'${ACTION_CODES_STRIKE_OFF_NOTICE}'.split(',')}")
+    private List<Long> strikeOffNoticeActionCodes;
+
+    @Autowired
+    private ApiLogger apiLogger;
+
+    @Bean
+    public List<ValidationRule<Long>> getActionCodeValidationRules() {
```
ICoxCH, line 32: Very minor but this may be better as a Set rather than a list? Unless we need them ordered? Happy to aprove regardless

ICoxCH (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/config/ValidationConfig.java
markpit: we do need them ordered, as we may need one rule to run before others. There is also a test for the ordering in the unit test.

PR #58, opened 2020-08-27
OBJ-226 Get action code from Oracle Query API

mwestacott:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -61,22 +64,28 @@ public ObjectionService(ObjectionRepository objectionRepository,
...
 
     @Override
     public Objection createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails) {
         Map<String, Object> logMap = buildLogMap(companyNumber, null, null);
         logger.infoContext(requestId, "Creating objection", logMap);
 
-        // TODO OBJ-231 process query result and return eligibility status
+        long actionCode = oracleQueryClient.getCompanyActionCode(companyNumber);
+
+        logger.debugContext(requestId, "Company action code is " + actionCode);
```
mwestacott, line 86: Since the story doesn't specify how to deal with negative scenarios (and the fact that @markpit 's PR that will do the action code checks is close to being finished) I think just logging the code might be fine for now.

ICoxCH:
I am happy with this, but this pr seems a good place. In the pom.xml there is a commented section about the jdbc and oracle driver dependencies. I believe with this approach using the query api, we can remove that commented section as we do not need it

ICoxCH (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -61,22 +64,28 @@ public ObjectionService(ObjectionRepository objectionRepository,
...
+        this.oracleQueryClient = oracleQueryClient;
     }
 
     @Override
     public Objection createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails) {
         Map<String, Object> logMap = buildLogMap(companyNumber, null, null);
         logger.infoContext(requestId, "Creating objection", logMap);
 
-        // TODO OBJ-231 process query result and return eligibility status
+        long actionCode = oracleQueryClient.getCompanyActionCode(companyNumber);
```
markpit, line 84: how about having the client call and logging in a separate method?
then we could just say
long actionCode = getActionCode(companyNumber);

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/client/OracleQueryClient.java
```diff
@@ -0,0 +1,25 @@
...
+    @Autowired
+    private RestTemplate restTemplate;
+
+    @Value("${ORACLE_QUERY_API_URL}")
+    private String oracleQueryApiUrl;
+
+    public Long getCompanyActionCode(String companyNumber) {
+        String getCompanyActionCodeUrl = String.format("%s/company/%s/action-code", oracleQueryApiUrl, companyNumber);
+
+        ResponseEntity<Long> response = restTemplate.getForEntity(getCompanyActionCodeUrl, Long.class);
```
markpit, line 21: what happens if we get a 404 returned?

mwestacott: The service errors - it's caught and logged by the controller and an internal server error is thrown. But, that's what lead me on to that issue we have (I think) within the web client, where it doesn't handle these correctly. It's not related to this change but a general problem. Will post something in Slack and then raise a defect.

This PR is fine though.

markpit (approved):

PR #59, opened 2020-08-28
OBJ-242 chips client stub

mattch1:
Stubbed client and unit test. As the method is void there is not a lot to add.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsClient.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsClient.java
@@ -4,9 +4,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import uk.gov.companieshouse.api.strikeoffobjections.common.ApiLogger;
 
-import java.util.HashMap;
-import java.util.Map;
-
 @Component
 public class ChipsClient {
 
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsClient.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsClient.java
@@ -19,7 +19,6 @@ public class ChipsClient {
 
     public void sendToChips(String requestId){
         // TODO OBJ-240 add contact model object and send to CHIPS
-        Map<String, Object> logMap = new HashMap<>();
-        apiLogger.infoContext(requestId, "Sending contact to CHIPS", logMap);
+        apiLogger.infoContext(requestId, "Sending contact to CHIPS");
     }
 }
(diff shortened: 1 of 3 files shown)


src/test/java/chips/ChipsClientTest.java
```diff
@@ -0,0 +1,33 @@
+package chips;
```
mwestacott, line 1: Package is wrong here. Needs to match where the class is and the ChipsClient import statement below should then not be needed.

src/test/java/chips/ChipsClientTest.java
```diff
@@ -0,0 +1,33 @@
...
+    private ChipsClient chipsClient;
+
+    @Mock
+    private ApiLogger apiLogger;
+
+    @Test
+    void testSendMessageToChips(){
+        // TODO OBJ-240 test sending contact data to CHIPS
+        chipsClient.sendToChips(REQUEST_ID);
+        assertEquals("test123", REQUEST_ID);
```
mwestacott, line 30: This assertion is pretty pointless - it's checking the unit-test code not the implementation. Appreciate there's very little at the moment in the implementation to check. Maybe this test class just isn't required at this point?

markpit: the real method will call apiLogger, so you could verify that apiLogger is called...

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsClient.java
```diff
@@ -0,0 +1,25 @@
...
+    private ApiLogger apiLogger;
+
+    @Autowired
+    public ChipsClient(ApiLogger apiLogger) {
+        this.apiLogger = apiLogger;
+    }
+
+    public void sendToChips(String requestId){
+        // TODO OBJ-240 add contact model object and send to CHIPS
+        Map<String, Object> logMap = new HashMap<>();
```
markpit, line 22: there should be an infoContext method without a logMap, no point passing in an empty one.

src/test/java/chips/ChipsClientTest.java
mattch1: Refactored

src/test/java/chips/ChipsClientTest.java
mattch1: Made assertion more meaningful - please bear mind that this is a stub.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsClient.java
mattch1: Removed empty log map.

markpit (approved):

markpit:
There's a couple of Sonar issues to fix

markpit (approved):

mwestacott (approved):

PR #61, opened 2020-08-28
OBJ-231 Integrate action code validation into create objection

markpit:
Plugging in the actionCodeValidator into ObjectionService so we can validate the action code when creating a new Objection

Also, the ObjectionService constructor args list had grown too large (Sonar) so had to move them out into fields.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
@@ -19,7 +19,7 @@ public class Objection {
         private String companyNumber;
         private String reason;
         private ObjectionStatus status;
-        private int actionCode;
+        private Long actionCode;
         private String httpRequestId;
 
         public Builder withCreatedOn(LocalDateTime createdOn) {
@@ -48,7 +48,7 @@ public class Objection {
             return this;
         }
 
-        public Builder withActionCode(int actionCode) {
+        public Builder withActionCode(Long actionCode) {
             this.actionCode = actionCode;
             return this;
         }
@@ -86,7 +86,7 @@ public class Objection {
     @Field("status")
     private ObjectionStatus status;
     @Field("action_code")
-    private int actionCode;
+    private Long actionCode;
     @Field("http_request_id")
     @JsonIgnore
     private String httpRequestId;
@@ -153,11 +153,11 @@ public class Objection {
         this.status = status;
     }
 
-    public int getActionCode() {
+    public Long getActionCode() {
         return actionCode;
     }
 
-    public void setActionCode(int actionCode) {
+    public void setActionCode(Long actionCode) {
         this.actionCode = actionCode;
     }
 
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
@@ -50,36 +50,32 @@ public class ObjectionService implements IObjectionService {
     private static final String ATTACHMENT_NOT_DELETED_SHORT = "Unable to delete attachment %s";
     private static final String INVALID_PATCH_STATUS = "Unable to patch status to %s for Objection id: %s";
 
+    @Autowired
     private ObjectionRepository objectionRepository;
-    private ApiLogger logger;
-    private Supplier<LocalDateTime> dateTimeSupplier;
-    private ObjectionPatcher objectionPatcher;
-    private FileTransferApiClient fileTransferApiClient;
-    private ERICHeaderParser ericHeaderParser;
-    private ObjectionProcessor objectionProcessor;
-    private OracleQueryClient oracleQueryClient;
-    private ActionCodeValidator actionCodeValidator;
 
     @Autowired
-    public ObjectionService(ObjectionRepository objectionRepository,
-                            ApiLogger logger,
-                            Supplier<LocalDateTime> dateTimeSupplier,
-                            ObjectionPatcher objectionPatcher,
-                            FileTransferApiClient fileTransferApiClient,
-                            ERICHeaderParser ericHeaderParser,
-                            ObjectionProcessor objectionProcessor,
-                            OracleQueryClient oracleQueryClient,
-                            ActionCodeValidator actionCodeValidator) {
-        this.objectionRepository = objectionRepository;
-        this.logger = logger;
-        this.dateTimeSupplier = dateTimeSupplier;
-        this.objectionPatcher = objectionPatcher;
-        this.fileTransferApiClient = fileTransferApiClient;
-        this.ericHeaderParser = ericHeaderParser;
-        this.objectionProcessor = objectionProcessor;
-        this.oracleQueryClient = oracleQueryClient;
-        this.actionCodeValidator = actionCodeValidator;
-    }
+    private ApiLogger logger;
+
+    @Autowired
+    private Supplier<LocalDateTime> dateTimeSupplier;
+
+    @Autowired
+    private ObjectionPatcher objectionPatcher;
+
+    @Autowired
+    private FileTransferApiClient fileTransferApiClient;
+
+    @Autowired
+    private ERICHeaderParser ericHeaderParser;
+
+    @Autowired
+    private ObjectionProcessor objectionProcessor;
+
+    @Autowired
+    private OracleQueryClient oracleQueryClient;
+
+    @Autowired
+    private ActionCodeValidator actionCodeValidator;
 
     @Override
     public Objection createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails) {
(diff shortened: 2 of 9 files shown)


ICoxCH (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
```diff
@@ -86,7 +86,7 @@ public Objection build() {
     @Field("status")
     private ObjectionStatus status;
     @Field("action_code")
-    private int actionCode;
+    private Long actionCode;
```
mattch1, line 89: Should this be primitive in the object itself?

markpit: shouldn't matter due to auto boxing, just that i'd done the validation with Long so tried to keep it consistent.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
```diff
@@ -95,37 +102,78 @@
...
         assertEquals(MOCKED_TIME_STAMP, objectionResponse.getCreatedOn());
         assertEquals(COMPANY_NUMBER, objectionResponse.getCompanyNumber());
-        assertEquals(ObjectionStatus.OPEN, objectionResponse.getStatus());
+        assertEquals(OPEN, objectionResponse.getStatus());
         assertEquals(AUTH_ID, objectionResponse.getCreatedBy().getId());
         assertEquals(E_MAIL, objectionResponse.getCreatedBy().getEmail());
     }
 
+    @Test
+    void createObjectionIneligibleStatusTest() throws ValidationException {
```
mattch1, line 154: Is there a test scenario for 209 ticket where the code is neither eligible nor ineligible - is that covered anywhere on another ticket perhaps?

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
```diff
@@ -69,6 +74,8 @@
     private static final String ATTACHMENT_ID = "12348765";
     private static final String REASON = "REASON";
     private static final String ACCESS_URL = "/dummyUrl";
+    private static final Long ACTION_CODE_OK = 3000L;
```
mattch1, line 77: Are these made up action/ineligibility codes for test purposes?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
mattch1: As long as there's no danger of Long being null that should be fine.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
markpit: yes, it's just to test that the same value is being populated in the objection that is saved to mongo.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
markpit: I hadn't thought of a null situation. I'll change to primitive then.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionServiceTest.java
markpit: The AllowedValuesValidationRuleTest should cover that scenario, which basically checks that if a value is in the list of 'ok' values, it is ok, if it is not in the list then a validation exception gets thrown.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
ICoxCH: What would be better a null pointer exception or a default value of 0 being assigned?

markpit: actually, yes, a null in mongo would indicate a missing action code, perhaps companies can have no action code anyway? 0 might indicate an actual 0 code or possible missing code. I think i'll leave it as is then.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/validation/AllowedValuesValidationRule.java
```diff
@@ -22,7 +22,7 @@ public AllowedValuesValidationRule(List<T> allowableValues,
     @Override
     public void validate(T input, String logContext) throws ValidationException {
         if (!allowableValues.contains(input)) {
-            apiLogger.debugContext(logContext, String.format("%s is not an allowed value", input));
+            apiLogger.debugContext(logContext, String.format("AllowedValuesValidationRule %s is not an allowed value", input));
```
mwestacott, line 25: You could use Class.getSimpleName() here instead of repeating the class name. Same in the other rule log statement.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
mattch1: We could have -1 - just think of what we return to the web module in this scenario.

markpit: I guess you could still have a -1 action code in chips DB

mattch1: As long the web will handle a value of null we should be ok.

markpit: thinking about it, we don't actually return the action code to the web, just the objection status (enum value)

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -94,12 +93,36 @@ public Objection createObjection(String requestId, String companyNumber, String
...
+
+        logger.debugContext(requestId, "Company action code is " + actionCode);
+
+        return actionCode;
+    }
+
+    private ObjectionStatus getObjectionStatusForCreate(Long actionCode, String companyNumber, String logContext) {
+        ObjectionStatus objectionStatus = ObjectionStatus.OPEN;
+        try {
+            actionCodeValidator.validate(actionCode, logContext);
```
mattch1, line 114: Might be a good idea to have a null check here if the action code is to be Long.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
mattch1: of course this action code does not get returned to the web - think of the wrong place. However, we may want to check for null where this action code intersects with creating the status that we do return to the web - added another comment below.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mwestacott: Maybe I shouldn't have used a non-primitive type in my earlier PR to get the action code, thereby making it clear (enforcing) that non-null values are just not allowed?

markpit: > Might be a good idea to have a null check here if the action code is to be Long.

if it is null, it should be validated as normal. The AllowedValuesValidationRule will throw a ValidationException containing ObjectionStatus.INELIGIBLE_NO_DISSOLUTION_ACTION if it is null, as null is not one of the codes specified in OBJ-209 (these codes are in config ACTION_CODES_STRIKE_OFF_NOTICE). I can add an extra unit test for that rule to test a null value being passed into it.

markpit: > Maybe I shouldn't have used a non-primitive type in my earlier PR to get the action code, thereby making it clear (enforcing) that non-null values are just not allowed?

I think it's ok, it covers a scenario where a company has no action code

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/validation/AllowedValuesValidationRule.java
markpit: updated to use getSimpleName()

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
markpit: I've added extra unit tests to the validation rules and they work as expected when passing null into them.

mattch1 (approved):
Good to have those extra unit tests.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mwestacott: > > Maybe I shouldn't have used a non-primitive type in my earlier PR to get the action code, thereby making it clear (enforcing) that non-null values are just not allowed?
> 
> I think it's ok, it covers a scenario where a company has no action code

Oh yes, true. And just seen that it is a nullable column in the database (although I did think that a company always had an action code).

mwestacott (approved):

PR #63, opened 2020-09-02
OBJ-241 Call through to CHIPS from the Objection Processor

mwestacott:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -70,7 +75,7 @@ public void process(Objection objection, String httpRequestId)
 
         // TODO update status to PROCESSING
 
-        // TODO OBJ-139/OBJ-20 do chips sending
+        chipsService.sendObjection(httpRequestId, objection);
```
markpit, line 78: I'd like to catch any exceptions thrown from chips service so we can log them, don't want to have to rely on other classes to do the logging. We can re-throw the exception (I know there's a log/throw anti pattern but I'd like to make sure nothing slips through).

mwestacott: Ok. Will implement as we've done for the e-mail calls.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ChipsService.java
```diff
@@ -0,0 +1,29 @@
...
+public class ChipsService implements IChipsService {
+
+    private final ChipsClient chipsClient;
+
+    @Autowired
+    public ChipsService(ChipsClient chipsClient) {
+        this.chipsClient = chipsClient;
+    }
+
+    public void sendObjection(String requestId, Objection objection) {
```
markpit, line 21: you can stick a @Overrides on here

markpit (approved):

ICoxCH (approved):

PR #66, opened 2020-09-08
Feature/obj 255 add name info to mongo email

mattch1:
Added user name and shareIdentity flag to create method - pass to mongo and email.

Added extra body parameter to the create method - had to do some refacorting in the service layer to prevent the parameter list from becoming too big.


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -134,7 +136,8 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
             @PathVariable("companyNumber") String companyNumber,
             @RequestHeader(value = ERIC_REQUEST_ID) String requestId,
             @RequestHeader(value = ERIC_IDENTITY) String ericUserId,
-            @RequestHeader(value = ERIC_AUTHORISED_USER) String ericUserDetails
+            @RequestHeader(value = ERIC_AUTHORISED_USER) String ericUserDetails,
+            @RequestBody ObjectionCreate objectionCreate
```
ICoxCH, line 140: Very minor I know, but as this change will modify the service interface, it is probably a good a time as any to bump the Major version as any apps built against the old interface will no longer work as expected

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IObjectionService.java
```diff
@@ -16,7 +18,13 @@
 import javax.servlet.http.HttpServletResponse;
 
 public interface IObjectionService {
-    Objection createObjection(String requestId, String companyNumber, String ericUserId, String ericUserDetails);
+    Objection createObjection(String requestId,
+                              String companyNumber,
+                              ObjectionUserDetails userDetails);
+
+    ObjectionUserDetails buildUserDetails(String ericUserId,
```
ICoxCH, line 25: Personally I would place this in a different (maybe new) service as it doesn't really have anything to do with an Objection.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
```diff
@@ -90,6 +103,10 @@ public Objection build() {
     @Field("http_request_id")
     @JsonIgnore
     private String httpRequestId;
+    @Field("full_name")
+    private String fullName;
+    @Field("share_identity")
+    private boolean shareIdentity;
```
ICoxCH, line 109: I would set this as a Boolean, so if the value is somehow not present in the db, we get null instead of false

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IObjectionService.java
mattch1: This is part of the objection -I'm not sure we would be getting anything better by doing that.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
mattch1: What we want here is probably a default of false.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: Will do.

mattch1: Changed version from 0.1 to 0.2

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -146,7 +149,15 @@ public ObjectionController(PluggableResponseEntityFactory responseEntityFactory,
         );
 
         try {
-            Objection objection = objectionService.createObjection(requestId, companyNumber, ericUserId, ericUserDetails);
+            ObjectionUserDetails userDetails = objectionService.buildUserDetails(
```
mwestacott, line 152: Kind of related to @ICoxCH 's comment below, but why not just pass the 'objectionCreate' object into the existing 'createObjection()' method and let it do what it needs to (in a private service method) to build an object to save in Mongo?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IObjectionService.java
mwestacott: Please see my comment higher up. I don't think we need this new public service method.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: It's undesirable because we end up with a method with a parameter list of five.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
```diff
@@ -21,6 +21,8 @@
         private ObjectionStatus status;
         private Long actionCode;
         private String httpRequestId;
+        private String fullName;
```
mwestacott, line 24: In terms of Mongo, should 'fullName' and 'shareIdentity' maybe live in the 'CreatedBy' bucket/object?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -78,28 +80,45 @@
...
-
         Objection entity = new Objection.Builder()
                 .withCompanyNumber(companyNumber)
                 .withCreatedOn(dateTimeSupplier.get())
-                .withCreatedBy(new CreatedBy(ericUserId, userEmailAddress))
+                .withCreatedBy(userDetails.getCreatedby())
                 .withHttpRequestId(requestId)
                 .withActionCode(actionCode)
                 .withStatus(objectionStatus)
+                .withFullName(userDetails.getFullName())
```
mwestacott, line 100: So if 'fullName' and 'shareIdentity' lived alongside 'email' in the created_by bucket, these lines wouldn't be needed.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IObjectionService.java
mattch1: Unfortineately we do need as we have to test this and we have to pass the data in somehow.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
markpit: That sounds better, have the fullName and shareIdentity in CreatedBy as it is user identity related

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mwestacott: Fair enough. Maybe just have 'buildUserDetails()' as a private helper method in this controller class then?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
mwestacott: Yep. It all sounds related.

mattch1: Put the new fields into createdBy deleted the user details class

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/create/ObjectionUserDetails.java
```diff
@@ -0,0 +1,43 @@
+package uk.gov.companieshouse.api.strikeoffobjections.model.create;
+
+import uk.gov.companieshouse.api.strikeoffobjections.model.entity.CreatedBy;
+
+public class ObjectionUserDetails {
+
+    private String fullName;
+    private boolean shareIdentity;
+    private String userEmailAddress;
```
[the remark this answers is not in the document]
mattch1: Removed this class.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
mattch1: Removed

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
mattch1: will live with the extra parameter.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
```diff
@@ -141,7 +140,7 @@ public void addAttachment(Attachment attachment) {
         return attachments;
     }
 
-    public void setAttachments(List<Attachment> attachments) {
+     public void setAttachments(List<Attachment> attachments) {
```
mwestacott, line 143: Errant space needs removing.

mattch1:
Just wondering if ObjectionCreate is the right name - torn between something like UserDetails which does not convey that it is part of the creation request body or the current name which does not convey that we are sending user details. I am just thinking about this in relation to the web ticket and naming it in there - we should keep the name consistent.

mwestacott:
> Just wondering if ObjectionCreate is the right name - torn between something like UserDetails which does not convey that it is part of the creation request body or the current name which does not convey that we are sending user details. I am just thinking about this in relation to the web ticket and naming it in there - we should keep the name consistent.

I thought that ObjectionCreate was ok since it was consistent with ObjectionPatch which is the class name used in the controller 'patch' method. Worth getting some more opinions on this though.

markpit:
> > Just wondering if ObjectionCreate is the right name - torn between something like UserDetails which does not convey that it is part of the creation request body or the current name which does not convey that we are sending user details. I am just thinking about this in relation to the web ticket and naming it in there - we should keep the name consistent.
> 
> I thought that ObjectionCreate was ok since it was consistent with ObjectionPatch which is the class name used in the controller 'patch' method. Worth getting some more opinions on this though.

I don't mind too much, but how about ObjectionCreateData or ObjectionCreateDetails, as I guess we are saying this is the data to create the objection with.

markpit:
> > > Just wondering if ObjectionCreate is the right name - torn between something like UserDetails which does not convey that it is part of the creation request body or the current name which does not convey that we are sending user details. I am just thinking about this in relation to the web ticket and naming it in there - we should keep the name consistent.
> > 
> > 
> > I thought that ObjectionCreate was ok since it was consistent with ObjectionPatch which is the class name used in the controller 'patch' method. Worth getting some more opinions on this though.
> 
> I don't mind too much, but how about ObjectionCreateData or ObjectionCreateDetails, as I guess we are saying this is the data to create the objection with.

saying that, it's up to you, I don't mind just ObjectionCreate

mwestacott:
> > > Just wondering if ObjectionCreate is the right name - torn between something like UserDetails which does not convey that it is part of the creation request body or the current name which does not convey that we are sending user details. I am just thinking about this in relation to the web ticket and naming it in there - we should keep the name consistent.
> > 
> > 
> > I thought that ObjectionCreate was ok since it was consistent with ObjectionPatch which is the class name used in the controller 'patch' method. Worth getting some more opinions on this though.
> 
> I don't mind too much, but how about ObjectionCreateData or ObjectionCreateDetails, as I guess we are saying this is the data to create the objection with.

We didn't add 'Data' or 'Details' to the end of the 'ObjectionPatch' class name but if we do change this, I think we need to in order to remain consistent.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
```diff
@@ -87,18 +90,33 @@
...
         objection.setId(OBJECTION_ID);
         objection.setStatus(ObjectionStatus.SUBMITTED);
-        when(objectionService.createObjection(REQUEST_ID, COMPANY_NUMBER, AUTH_ID, AUTH_USER)).thenReturn(objection);
+
+        when(objectionService.createObjection(
+                eq(REQUEST_ID),
+                eq(COMPANY_NUMBER),
+                eq(AUTH_ID),
+                eq(AUTH_USER),
+                any()
```
markpit, line 105: instead of any() you should use the objectionCreate that is created above that will be passed in as a param

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
```diff
@@ -111,12 +129,17 @@ void createObjectionTest() {
...
     void createObjectionEligibilityErrorTest() {
+        ObjectionCreate objectionCreate = new ObjectionCreate();
+        objectionCreate.setFullName(FULL_NAME);
+        objectionCreate.setShareIdentity(false);
         Objection objection = new Objection();
         objection.setId(OBJECTION_ID);
         objection.setStatus(ObjectionStatus.INELIGIBLE_COMPANY_STRUCK_OFF);
-        when(objectionService.createObjection(REQUEST_ID, COMPANY_NUMBER, AUTH_ID, AUTH_USER)).thenReturn(objection);
+        when(objectionService.createObjection(eq(REQUEST_ID), eq(COMPANY_NUMBER),
+                eq(AUTH_ID), eq(AUTH_USER), any())).thenReturn(objection);
```
markpit, line 139: same here, you can replace any() with the expected objectionCreate

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
```diff
@@ -127,13 +150,18 @@ void createObjectionEligibilityErrorTest() {
 
     @Test
     void createObjectionExceptionTest() {
-        when(objectionService.createObjection(REQUEST_ID, COMPANY_NUMBER, AUTH_ID, AUTH_USER))
+        ObjectionCreate objectionCreate = new ObjectionCreate();
+        objectionCreate.setFullName(FULL_NAME);
+        objectionCreate.setShareIdentity(false);
+        when(objectionService.createObjection(eq(REQUEST_ID), eq(COMPANY_NUMBER),
+                eq(AUTH_ID), eq(AUTH_USER), any()))
```
markpit, line 157: same here, you can replace any() with the expected objectionCreate

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
mattch1: Updated

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
mattch1: Updated

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionControllerTest.java
mattch1: Updated.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
mattch1: Removed

mwestacott (approved):

markpit (approved):

PR #72, opened 2020-09-28
Feature/obj 297 user details in patch

mattch1:
API changes relating to OBJ-284 so that when "change" is selected for user details and continue is pressed the details are sent to the api which needs to pick them up and process them along with any other changed data.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/CreatedBy.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/CreatedBy.java
@@ -35,7 +35,15 @@ public class CreatedBy {
         return fullName;
     }
 
+    public void setFullName(String fullName) {
+        this.fullName = fullName;
+    }
+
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/CreatedBy.java
```diff
@@ -35,7 +35,15 @@ public String getFullName() {
         return fullName;
     }
 
+    public void setFullName(String fullName) {
```
ICoxCH, line 38: Do we need setters for these? id and email don't have them, if possible we should leave created by as an immutable object

mattch1: Yes we'll want them - the patcher needs them. Not sure the price is right for keeping it immutable we'd have to create a new object for each field in te patcher.

ICoxCH (approved):

markpit (approved):

PR #73, opened 2020-09-29
OBJ-299 annotations to receiver name format correction in the web

mattch1:
Just updated the two PRs - the created by returned when GET is called for display on the check your answers page was left as it was the thing causing problems with serialization earlier - but if I add a DTO for created by then the var can be reformatted here as well.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/CreatedByResponseDTO.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/CreatedByResponseDTO.java
@@ -1,0 +1,56 @@
+package uk.gov.companieshouse.api.strikeoffobjections.model.response;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CreatedByResponseDTO {
+
+    @JsonProperty("id")
+    private String id;
+
+    @JsonProperty("email")
+    private String email;
+
+    @JsonProperty("full_name")
+    private String fullName;
+
+    @JsonProperty("share_identity")
+    private boolean shareIdentity;
+
+    public CreatedByResponseDTO() {
+
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
+
+    public String getFullName() {
+        return fullName;
+    }
+
+    public void setFullName(String fullName) {
+        this.fullName = fullName;
+    }
+
+    public boolean isShareIdentity() {
+        return shareIdentity;
+    }
+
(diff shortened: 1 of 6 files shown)


markpit (approved):

mwestacott (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/ObjectionResponseDTO.java
```diff
@@ -17,7 +17,7 @@
     private String createdOn;
 
     @JsonProperty("created_by")
-    private CreatedBy createdBy;
+    private CreatedByResponseDTO createdBy;
```
markpit, line 20: unused import for CreatedBy

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/CreatedByResponseDTO.java
```diff
@@ -0,0 +1,56 @@
...
+    @JsonProperty("email")
+    private String email;
+
+    @JsonProperty("full_name")
+    private String fullName;
+
+    @JsonProperty("share_identity")
+    private boolean shareIdentity;
+
+    public CreatedByResponseDTO() {
```
markpit, line 21: The SonarLint IntelliJ plugin has highlighted the empty constructor for No empty methods allowed. Classes will have an empty public constructor by default if you don't include one.

markpit:
I'm not up to speed with what the actual problem is, but if it is a serialization issue, it might be just that CreatedBy needs to implement the Serializable interface.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/CreatedByResponseDTO.java
markpit: Could also remove the empty constructor from ObjectionResponseDTO while you're at it

mattch1:
No the serialisation problem was due to a lack of Json properties for the camel case fields - so introduced the DTO.

mwestacott:
Right, was struggling, but think I understand now why you've made the additional changes @mattch1 . The CreatedBy 'entity' object was being returned in the response DTO which didn't (and shouldn't) have the `JsonProperty` annotations on to change the field name to have underscores. So a new DTO was required and a corresponding mapper for going from the entity to the new DTO.

I'm happy with this PR but suggest you address @markpit 's comments regarding the constructors before merging.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/CreatedByResponseDTO.java
mattch1: Removed constructor for CreatedByResponseDTO, the one for the ObjectionResponseDTO is being used by the POST method.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/ObjectionResponseDTO.java
mattch1: Gone.

markpit (approved):

PR #75, opened 2020-10-02
OBJ-312 Adding null checks for attachments in ChipsRequest

markpit:
Made the new attachments mapping null safe and found a neat way of converting the attachments map to a string (also null safe).
Added tests to cover these.

### DIFF
--- a/src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/chips/ChipsRequestTest.java
+++ b/src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/chips/ChipsRequestTest.java
@@ -8,6 +8,8 @@ import java.util.ArrayList;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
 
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/chips/ChipsRequestTest.java
```diff
@@ -42,4 +42,43 @@ void testFieldsAndAttachmentConstruction()
...
+                CUSTOMER_EMAIL,
+                REASON,
+                DOWNLOAD_URL_PREFIX
+        );
+
+        // should not throw null pointer exception
+        assertNull(chipsRequest.getAttachments());
+    }
+
+    @Test
```
mwestacott, line 62: I'm curious as to what gets output when the attachments are 'null'. I'm guessing 'null' features in the output string but worth checking/proving this with another unit-test I think.

markpit: attachments=''. - I've now added a test to cover it

mwestacott (approved):

PR #76, opened 2020-10-08
Feature/obj 322 add eligibility endpoint

mattch1:
Added endpoint and new eligibility object as per slack discussion. Unit tested but have yet to get it working due to required changes in the interceptors which are still causing problems. At present nothing is calling this endpoint - anything introduced that does will need to be feature flagged while the interceptor problem is being handled.

As the error is triggered on the endpoint call and not reaching the code the feature flag cannot be added here.

The interceptor work will be completed on another ticket - although only one line needs to be changed there will be side effects of such a change and they will need to be handled as well.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
@@ -22,6 +22,7 @@ import uk.gov.companieshouse.api.strikeoffobjections.exception.AttachmentNotFoun
 import uk.gov.companieshouse.api.strikeoffobjections.exception.ObjectionNotFoundException;
 import uk.gov.companieshouse.api.strikeoffobjections.file.FileTransferApiClientResponse;
 import uk.gov.companieshouse.api.strikeoffobjections.model.create.ObjectionCreate;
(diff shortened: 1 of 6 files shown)


mwestacott:
Have you added a JIRA sub-task for the interceptor change that's needed or are you intending to do it under the OBJ-322 ticket (but in a separate PR)?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/controller/ObjectionController.java
```diff
@@ -501,4 +502,11 @@ public ResponseEntity deleteAttachment(
             );
         }
     }
+
+    @GetMapping("/eligibility")
```
mwestacott, line 506: Spec needs updating to align to this new end-point name and what's returned. @SVarallo might be doing this? Strictly speaking, should be done before this code is merged.

SVarallo: PR for this at: https://github.com/companieshouse/strike-off-objections-api/pull/77

markpit (approved):

mattch1:
Subtask OBJ-337 created for the interceptor work.

mwestacott (approved):

PR #78, opened 2020-10-10
OBJ-322 Removing use of ObjectionStatus in company eligible check

markpit:
Changing company eligible check to just use the actionCodeValidator and remove the need for converting to an ObjectionStatus to get the result.
Refactoring calls to actionCodeValidator so we only have to log the action code once

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
@@ -124,27 +124,35 @@ public class ObjectionService implements IObjectionService {
     private ObjectionStatus getObjectionStatusForCreate(Long actionCode, String companyNumber, String logContext) {
         ObjectionStatus objectionStatus = ObjectionStatus.OPEN;
         try {
-            actionCodeValidator.validate(actionCode, logContext);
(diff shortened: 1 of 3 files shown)


mwestacott (approved):

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -124,25 +124,35 @@ private Long getActionCode(String companyNumber, String requestId) {
     private ObjectionStatus getObjectionStatusForCreate(Long actionCode, String companyNumber, String logContext) {
         ObjectionStatus objectionStatus = ObjectionStatus.OPEN;
         try {
-            actionCodeValidator.validate(actionCode, logContext);
+            validateActionCode(actionCode, companyNumber, logContext);
```
mattch1, line 127: We probably don't need this method anymore we could call validateActionCode from the createObjection method as we are now doing from isEligible.

markpit: it was split out to improve readability of the createObjection method. If we merge them together it will make the createObjection longer and less readable.

mattch1 (approved):

PR #79, opened 2020-10-12
Feature/obj 337 interception handling for eligibility

mattch1:
Excluded the eligibility endpoint from interceptor checks.


markpit (approved):

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/EligibilityIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private ObjectionInterceptor objectionInterceptor;
+
+    @MockBean
+    private CompanyNumberInterceptor companyNumberInterceptor;
+
+    @MockBean
+    private UserAuthorizationInterceptor userAuthorizationInterceptor;
+
+    @Test
+    void willAlwaysAllowEligibilityEndpoint() throws Exception {
```
mwestacott, line 66: A slight improvement on this method name would be to mention that the interceptors aren't called, since that's the change you made (it could still not be allowed for other reasons, e.g. if X-Request-Id was missing from header). How about 'interceptorsNotCalledForEligibilityEndpoint'?

mattch1: Godd idea - changed the name.

mwestacott (approved):

PR #80, opened 2020-10-12
OBJ-20 Adding feature flag for sending chips contact data

markpit:
Adding new feature flag to allow us to not send the contact data to chips if chips code to handle it is not deployed.
The flag being set to true will mean that nulls get set in the ChipsRequest which has been updated to ignore null fields when converting to json.

### DIFF
--- a/src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/chips/ChipsRequestTest.java
+++ b/src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/chips/ChipsRequestTest.java
@@ -1,5 +1,8 @@
 package uk.gov.companieshouse.api.strikeoffobjections.model.chips;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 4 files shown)


markpit:
Sonar code smell is the existing TODO comment

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/model/chips/ChipsRequestTest.java
```diff
@@ -102,4 +107,28 @@ void testToStringNullAttachments() {
 
         assertEquals(expectedOutput, actualOutput);
     }
+
+    @Test
```
mwestacott, line 111: This unit test seems to be testing Jackson code/behaviour and not ours. Is it needed?

markpit: good spot, i left didn't need the extra line in the test. Removed now. The test will pass if the ChipsRequest has the @JsonInclude(JsonInclude.Include.NON_NULL)
annotation and fail if someone removes it.

ICoxCH (approved):

PR #88, opened 2020-10-14
SONAR - Removing Serializable from Attachment

markpit:
The links are being highlighted to be made serializable because Attachments is serializable (links are in Attachments) but Attachments is in Objection which is not Serializable. Removed Serializable from Attachments and tested email sending, mongo entries and chips contact, all seems ok.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionLinkKeys.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionLinkKeys.java
@@ -10,7 +10,7 @@ public enum ObjectionLinkKeys implements LinkKey {
 
     private final String key;
 
-    private ObjectionLinkKeys(String key) {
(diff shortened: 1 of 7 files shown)


markpit:
Ian raised a good point about raising a PR to make the original Links class serializable in it's own repo, so i'll look into that.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/ObjectionLinkKeys.java
```diff
@@ -10,7 +10,7 @@
 
     private final String key;
 
-    private ObjectionLinkKeys(String key) {
+    ObjectionLinkKeys(String key) {
```
markpit, line 13: IDE highlighted 'modifier private is redundant for Enum constructors'

mwestacott (approved):

ICoxCH (approved):

PR #89, opened 2020-10-15
obj-356 Updated model to include name and share identity

mattch1:
https://companieshouse.atlassian.net/browse/OBJ-356

Updated the ChipsRequest model  and service with full name and share identity.

Updated the chips service, request and client unit tests to include the two new fields.


mwestacott:
Change looks good to me, but if we merge it and it goes in to any environments where FEATURE_FLAG_SEND_CHIPS_CONTACT_DATA = true (so Toro1 and vagrant DEV, I think) it will break the CHIPS contact behaviour as the new fields will be unrecognised.

We'll either need to wait until the CHIPS changes are completed and deployed to those envs or we set the feature flag in those envs to 'false'.

ICoxCH:
Could you please merge master into this pr to bring in your pom file changes, so sonar can properly evaluate the pr?

markpit (approved):

mwestacott (approved):

PR #90, opened 2020-10-16
OBJ-162 generate reference number

ICoxCH:
On average can produce about 50 thousand a day before a collision. Seeing as CH recieves 40 thousand a year, a collision should be enough of an edge case scenario that regenerating an id in case of collision won't be a real issue


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IIdGeneratorService.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.companieshouse.api.strikeoffobjections.service;
+
+import java.security.NoSuchAlgorithmException;
+
+public interface IIdGeneratorService {
```
mwestacott, line 5: Wondering about the naming of this interface, method and other places in this PR. The story and business requirements all talk about a 'reference number' so should we adopt this terminology rather than 'id' (it sounds more like a system internal number that's hidden from the user and this isn't).

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IIdGeneratorService.java
```diff
@@ -0,0 +1,7 @@
...
+    String generateId() throws NoSuchAlgorithmException;
```
mwestacott, line 6: I can't see where the declared exception is thrown or declared in the actual implementing method. Is it required?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/IdGeneratorService.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.UUID;
+import java.util.function.Supplier;
+
+@Service
+public class IdGeneratorService implements IIdGeneratorService {
+
+    private static final String ID_PREFIX = "OBJ-";
```
mwestacott, line 15: So REFERENCE_NUMBER_PREFIX...?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IIdGeneratorService.java
ICoxCH: I was creating this with the thought that i'd be replacing the id of the mongo record with it, hence id.
I suppose this will need a group discussion, but if this "reference number" is a new field, then we will need to find how to add an index to a mongo collection in live, as this field definitely should be indexed.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/IdGeneratorService.java
```diff
@@ -0,0 +1,48 @@
...
+    @Autowired
+    public IdGeneratorService(Supplier<LocalDateTime> dateTimeSupplier) {
+        this.dateTimeSupplier = dateTimeSupplier;
+    }
+
+    @Override
+    public String generateId() {
+        LocalDate dateNow = dateTimeSupplier.get().toLocalDate();
+        String now = Long.toHexString(dateNow.toEpochDay());
+        String uuid = now + UUID.randomUUID().toString();
```
mwestacott: Var name here isn't strictly correct - it's 'now' + a UUID. Could we rename? Maybe just id would do?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IIdGeneratorService.java
mwestacott: Yes. Group discussion probably required. Indexing presumably needed if searching against the field, but will that happen or will CH staff look for reference number in their emails or in CHIPS. Bit unsure of the process here...

ICoxCH: the service will need to search against the reference to check its a unique reference before committing it to the database.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorService.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final Supplier<LocalDateTime> dateTimeSupplier;
+
+    @Autowired
+    public ReferenceNumberGeneratorService(Supplier<LocalDateTime> dateTimeSupplier) {
+        this.dateTimeSupplier = dateTimeSupplier;
+    }
+
+    @Override
+    public String generateReferenceNumber() {
```
mwestacott, line 29: I think some Javadoc with an example of the kind of reference number this method will generate would help interpret the code.

src/test/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorServiceTest.java
```diff
@@ -0,0 +1,35 @@
...
+    private ReferenceNumberGeneratorService referenceNumberGeneratorService;
+
+    @Test
+    void generateIdTest() {
+        when(supplier.get()).thenReturn(LocalDateTime.of(2020, Month.OCTOBER, 6, 9, 30));
+        String referenceNumber = referenceNumberGeneratorService.generateReferenceNumber();
+
+        assertEquals("OBJ-", referenceNumber.substring(0, 4));
+        assertEquals(18, referenceNumber.length());
+    }
```
mwestacott, line 34: I think you need a couple more tests, to check that only alphanumeric characters (letters in uppercase) and '-'s are present in the number. Also that the separators appear at the right place.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/IIdGeneratorService.java
mwestacott: Ah right. However.... we could just accept the risk of clashes is minimal and in that case, allow duplicates anyway and if helping a customer rely on some additional information to find the correct objection. Just an option worth exploring.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorService.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Autowired
+    public ReferenceNumberGeneratorService(Supplier<LocalDateTime> dateTimeSupplier) {
+        this.dateTimeSupplier = dateTimeSupplier;
+    }
+
+    /**
+     * Generates a 12 digit reference number based on the day and a random uuid, and inserts a hyphen every 4 characters
+     * e.g. OBJ-1F3C-A2E4-5D6B
+     * @return a 12 digit reference with hyphens every 4 digits, prefixed with REFERENCE_NUMBER_PREFIX
```
mwestacott, line 27: If you replace REFERENCE_NUMBER_PREFIX with {@value #REFERENCE_NUMBER_PREFIX} the generated Javadoc will handle it nicely (display constant name but link to actual value).

mwestacott (approved):
Minor comment related to Javadoc but approving.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorService.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Generates a 12 digit reference number based on the day and a random uuid, and inserts a hyphen every 4 characters
+     * e.g. OBJ-1F3C-A2E4-5D6B
+     * @return a 12 digit reference with hyphens every 4 digits, prefixed with {@value #REFERENCE_NUMBER_PREFIX}
+     */
+    @Override
+    public String generateReferenceNumber() {
+        LocalDate dateNow = dateTimeSupplier.get().toLocalDate();
+        String now = Long.toHexString(dateNow.toEpochDay());
```
markpit, line 32: You get less clashes if you remove the use of the date and just use the UUID. I created a test harness you can use here - https://github.com/companieshouse/strike-off-objections-api/compare/feature/OBJ-162-generate-user-friendly-reference-number...feature/ref-num-test-harness
Over a 5 year period i was getting about half a dozen clashes using the date approach, with UUID it was usually 0 or 1.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorService.java
```diff
@@ -0,0 +1,53 @@
...
+        unformattedReference = unformattedReference.substring(0, 12);
+        unformattedReference = unformattedReference.toUpperCase();
+
+        char separator = '-';
+
+        StringBuilder sb = new StringBuilder(unformattedReference);
+
+        for(int i = 0; i < unformattedReference.length() / 4 -1; i++) {
+            sb.insert(((i + 1) * 4) + i, separator);
+        }
```
markpit, line 49: the unformattedReference.length will change in every iteration, it works out ok as / acts as DIV so 14/4 will return the same as 12/4. This approach will also call length() each iteration so I would make calculation outside the loop.
I would have 
`int groupSize = 4;
  int numOfSeparatorsToAdd = (unformattedReference.length() / groupSize) - 1;
  StringBuilder sb = new StringBuilder(unformattedReference);

  for(int i = 0; i < numOfSeparatorsToAdd; i++) {
      sb.insert(((i + 1) * groupSize) + i, separator);
  }`

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorService.java
markpit: we could reduce the probability of clashes further by randomising the position of the separators, as long as they are followed by at least 1 char. The ref no. length would stay the same, might have to check with business if this is ok though.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorService.java
ICoxCH: I agree with moving the calculation outside of the loop, will do that.
But not sure unformattedReference.length would change every iteration.

markpit: ah yes, i was thinking it was the string buffer.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ReferenceNumberGeneratorService.java
ICoxCH: Have been doing some tests and it does seem we were trying to be to clever. Just a UUID stripped down to 12 characters seems less likely to cause a collision. I suppose at some point in the future prefixing the ID with the day would make collisions less likely as you know each day can't collide with a previous day. Whereas just a UUID any Id can collide with any id ever. But this seems like it would be greater than 10 years in the future with 5000 a day which is way more than we would ever need. So will just switch to just using UUID.

As for randomising the seperators, I don't think we should do that as the seperators aren't actually part of the id. More just there to break it up for a user reading it over the phone. See bullet point 2 in "Showing reference number to user" https://companieshouse.atlassian.net/browse/OBJ-150

markpit (approved):

PR #91, opened 2020-10-20
OBJ-169 replace mongo id with generated ref id

ICoxCH:


mwestacott:
I know it's more work, but it doesn't seem right to me that we are now just storing the reference number in the id field - it's a bit of business related data and not an internal DB id. You wouldn't do that in a relational database due to foreign key considerations but I suppose Mongo is a bit different. 

I know we need to be pragmatic so not too precious about this, but good to have the discussion...

ICoxCH:
> I know it's more work, but it doesn't seem right to me that we are now just storing the reference number in the id field - it's a bit of business related data and not an internal DB id. You wouldn't do that in a relational database due to foreign key considerations but I suppose Mongo is a bit different.
> 
> I know we need to be pragmatic so not too precious about this, but good to have the discussion...

Whilst in conversation with Kev and Steve about indexes and using this number, they both agreed that using this ref number as the db id would be the best approach as that way it gets indexed for free, and both the refId and ObjectId would of been Unique identifiers which didn't make sense to them.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -100,9 +105,15 @@ public Objection createObjection(String requestId,
                 .withActionCode(actionCode)
                 .withStatus(objectionStatus)
                 .withStatusChangedOn(dateTimeSupplier.get())
+                .withId(referenceNumberGeneratorService.generateReferenceNumber())
                 .build();
 
-        return objectionRepository.save(entity);
+        try {
+            return objectionRepository.insert(entity);
```
markpit, line 112: just wondering, should we have a pre defined number of retry attempts before exception-ing?

ICoxCH: I would argue that we should, but not as MVP, and would need its own story with acceptance criteria

markpit (approved):

mwestacott (approved):

PR #93, opened 2020-10-26
OBJ-233 Check for invalid REST requests to the Objections API

mwestacott:
Stop objections being submitted with missing user data and attempts to
change objection data after submission.


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
```diff
@@ -191,4 +192,10 @@ public LocalDateTime getStatusChangedOn() {
     public void setStatusChangedOn(LocalDateTime statusChangedOn) {
         this.statusChangedOn = statusChangedOn;
     }
+    
+    public boolean isDataEnteredByUserIncomplete() {
+        return StringUtils.isEmpty(reason)
```
mwestacott, line 197: I only implemented a check for the user entered data being incomplete here (since logic in web and API code should already mean the other fields can’t be empty) but maybe we should just check that the entire objection data is as we expect at this point?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessor.java
```diff
@@ -106,6 +108,18 @@ private void validateObjectionStatus(Objection objection, String httpRequestId)
         }
     }
 
+    private void validateObjectionData(Objection objection, String httpRequestId) throws ServiceException {
+        if (objection.isDataEnteredByUserIncomplete()) {
```
mwestacott, line 112: This addresses the main issue reported in OBJ-233 - fast-forwarding to the check your answers screen by changing the URL and submitting an Objection with incomplete data.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/ObjectionInterceptor.java
```diff
@@ -35,6 +37,20 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
 
         try {
             Objection objection = objectionService.getObjection(requestId, objectionId);
+            
+            // Operations on objections via the API REST interface are only allowed whilst the objection is
```
mwestacott, line 41: This is a ‘belt and braces’ change since removing the objection id from the web session once an objection is successfully submitted - see change in https://github.com/companieshouse/strike-off-objections-web/pull/155 - should ensure the API doesn’t receive any further REST calls for that objection. But if it did (e.g. by the API being called directly from an untrusted client) this check will stop them going through.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
mattch1: I think reason is the only field that could ever be empty here. It seems strange to check come mandatory fields but not others unless we can think of a possible edge case where they would be empty.

mattch1: Ignore this I can see how these might be null now.

mattch1:
I can see it would be tricky but I'm guessing you did not find a way of unit or integration testing this - no mock complete and incomplete datta we can send to trigger this exception.

ICoxCH:
It seems that there are unit test errors with this pr, not sure why this pr wasnt updated to let you know
https://ci.platform.aws.chdev.org/teams/team-toro-loco/pipelines/strike-off-objections-api/jobs/analyse-pull-request/builds/282

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/ObjectionInterceptor.java
markpit: what happens when we submit via Patch, won't the ObjectionStatus.OPEN != objection.getStatus() come in to play?

markpit: hold on, we submit when open, sorry.

markpit:
Approach looks ok to me

mwestacott:
> I can see it would be tricky but I'm guessing you did not find a way of unit or integration testing this - no mock complete and incomplete data we can send to trigger this exception.

No, didn't want to start unit or integration testing until we were happy with the approach. Positive comments so far so will start looking at the testing now and get the full reviews up once done.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/entity/Objection.java
```diff
@@ -191,4 +192,10 @@ public LocalDateTime getStatusChangedOn() {
...
+                || StringUtils.isEmpty(createdBy.getFullName())
+                || attachments.size() == 0;
```
ICoxCH, line 199: Sonar is recommending you use isEmpty() here

markpit (approved):

ICoxCH (approved):

PR #95, opened 2020-10-29
OBJ-381 return what spec says will be returned

ICoxCH:


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/eligibility/ObjectionEligibility.java
```diff
@@ -5,13 +5,13 @@
...
-    private boolean isEligible;
+    private boolean eligible;
 
-    public ObjectionEligibility(boolean isEligible) {
-        this.isEligible = isEligible;
+    public ObjectionEligibility(boolean eligible) {
+        this.eligible = eligible;
     }
 
     public boolean isEligible() {
```
ICoxCH, line 14: Jackson assumes this corresponds to a field called eligible, as there wasn't one it was adding an "eligible" field to the returned json.
Changing the field to eligible removes this issue

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/ObjectionResponseDTO.java
```diff
@@ -30,6 +33,10 @@
     @JsonProperty("status")
     private ObjectionStatus status;
 
+    @JsonProperty("links")
+    @JsonUnwrapped
+    private Links links;
```
ICoxCH, line 38: The spec mentions Objection has a links field, as the responseDTO is created using MapStruct, this was the best I could think of without more change, any suggestions welcome

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/eligibility/ObjectionEligibility.java
mwestacott: Just interested, how come this worked before? Didn't it mean we'd send a field back to the client that wasn't recognised or is there a change to do in 'web' too?

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/ObjectionResponseDTO.java
```diff
@@ -67,6 +75,7 @@ public String getCompanyNumber() {
 
     public void setCompanyNumber(String companyNumber) {
         this.companyNumber = companyNumber;
+        this.setLinks();
```
mwestacott, line 78: Doing this just doesn't look right. I'll have a look at code to understand the issue and see if I can see an alternative solution.

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/eligibility/ObjectionEligibility.java
ICoxCH: Jackson was sending two fields.
One it got from the JSON property annotation "is_eligible"
And another it got from inferring there should be a field called "eligible" from this getters name.

So the json response looked like
{
    is_eligible: true,
    eligible: true
}

I think the web was just ignoring the other field

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/response/ObjectionResponseDTO.java
ICoxCH: Ideally we'd be able to add a little private method to the objection mapper "generateDTOLinks"
Then call objectionResponseDTO.setLinks(generateDTOLinks(ObjectionEntity); as part of the mapping.

Bu that's all abstracted away into mapstruct, and not sure if there's a way we could do that

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/model/eligibility/ObjectionEligibility.java
mwestacott: Ah right 👍 .

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/service/impl/ObjectionService.java
```diff
@@ -283,11 +285,12 @@ public Attachment getAttachment(
         );
     }
 
-    private Links createLinks(String attachmentsUri, String attachmentId) {
-        String linkToSelf = attachmentsUri + "/" + attachmentId;
+    private Links createLinks(String selfLink, boolean download) {
```
mwestacott, line 288: Minor but could rename second var to 'createDownloadLink' which would improve readability in the 'if' statement lower down.

mwestacott:
Looking good. One minor spot, plus you have conflicts to deal with.

mwestacott (approved):

markpit (approved):

PR #96, opened 2020-11-02
OBJ-397 cannot download attachment of processed objection

markpit:
The Objection Status check in ObjectionInterceptor was blocking downloads due to the objection status not being OPEN.
I've extracted the status check into it's own interceptor to give us greater control over how we apply it.
We can now exclude the download path from the status check.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/CompanyNumberInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/CompanyNumberInterceptor.java
@@ -32,7 +32,7 @@ public class CompanyNumberInterceptor implements HandlerInterceptor {
         final boolean companyNumberMatches = doCompanyNumbersMatch(companyNumber, objection);
 
         if (!companyNumberMatches) {
-            apiLogger.errorContext(requestId, "Provided company number does not match objection company number", null);
+            apiLogger.infoContext(requestId, "Provided company number does not match objection company number");
             response.setStatus(HttpStatus.BAD_REQUEST.value());
         }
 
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/CompanyNumberInterceptor.java
```diff
@@ -32,7 +32,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
         final boolean companyNumberMatches = doCompanyNumbersMatch(companyNumber, objection);
 
         if (!companyNumberMatches) {
-            apiLogger.debugContext(requestId, "Provided company number does not match objection company number");
+            apiLogger.errorContext(requestId, "Provided company number does not match objection company number", null);
```
markpit, line 35: There was nothing logged when eg. company number was incorrect for url so I've updated the error condition logging to errorContext. This was already the case for the ObjectionStatus.

ICoxCH: Imo this isn't an error scenario? Maybe worth bumping it from debug to info, but I don't think this deserves an error log unless an exception being thrown

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/interceptor/ObjectionStatusInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws Exception {
+        final String requestId = request.getHeader(ERICHeaderFields.ERIC_REQUEST_ID);
+        apiLogger.debugContext(requestId, "Checking provided company number matches objection company number");
+        final Objection objection = (Objection) request.getAttribute(InterceptorConstants.OBJECTION_ATTRIBUTE);
+
+        // Operations on objections via the API REST interface are only allowed whilst the objection is
+        // still 'open', i.e. has not yet been submitted (to CHIPS) for processing
+        if (ObjectionStatus.OPEN != objection.getStatus()) {
+            apiLogger.errorContext(
```
ICoxCH, line 32: Similar to my other comment, I know this was existing behaviour, but does it need an error log? This is defined behaviour for the application, maybe an info log?

ICoxCH (approved):

SVarallo (approved):

PR #98, opened 2020-11-04
OBJ-408 move company profile call during processing

markpit:
Moving call to get company profile before sending chips data during processing to prevent half processed cases if company profile api is unavailable.


src/test/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessorTest.java
```diff
@@ -274,7 +299,7 @@ void processHandlesCustomerEmailUncheckedException() throws ServiceException {
     }
 
     @Test
-    void processHandlesChipsUncheckedException() throws ServiceException {
+    void processHandlesChipsUncheckedException() {
```
markpit, line 302: the throws was highlighted as not needed in IntelliJ

ICoxCH (approved):

PR #100, opened 2020-11-10
Added a dev tiltfile to repo

ICoxCH:


Tiltfile.dev
```diff
@@ -0,0 +1,18 @@
...
+  command = 'mvn compile jib:dockerBuild -Dimage=$EXPECTED_REF',
+  live_update = [
+    sync(
+      local_path = './target/classes',
+      remote_path = '/app/classes'
+    ),
+    restart_container()
+  ],
+  deps = ['./target/classes']
+)
```
mattch1, line 18: Newline.

ICoxCH: Happy to add a new line, but what does the newline provide? Why is it needed

mattch1: I think sonar complains about it; github always does.

mattch1 (approved):

Tiltfile.dev
ICoxCH: Sonar didn't raise it as a smell, bug, hotspot or vulnerability so I think Sonar is ok with it.
And the Red circle on Github is less complaining and more letting you know. I am yet to ever encounter an issue caused by this

mattch1: In that case it's just a styling preference rather than anything else.

PR #101, opened 2020-11-11
Marking servlet-api as provided to get app running in Docker and remo…

markpit:
…ved duplicate deps

Changing pom.xml to mark servlet-api as provided to get app running in Docker and removing spring boot dependencies that were in there twice.


ICoxCH (approved):
This works in my env as expected.
Removing the duplicate spring boot starters should definitely be done.

Having to put servlet-api as provided doesn't seem to have broken anything

PR #102, opened 2020-11-12
OBJ-413 Make api accept objections with empty reason text

markpit:
Removing a reason text check to allow objections with empty reasons to be processed.


src/test/java/uk/gov/companieshouse/api/strikeoffobjections/processor/ObjectionProcessorTest.java
```diff
@@ -129,13 +129,20 @@ void processThrowsIllegalArgumentExceptionTest() {
     }
 
     @Test
-    void processThrowsServiceExceptionIfUserDataIsMissingReason() {
+    void processShouldNotThrowServiceExceptionIfUserDataIsMissingReason() throws Exception {
```
markpit, line 132: updated unit test to make sure exception is not thrown if reason text empty.

mattch1 (approved):

PR #104, opened 2020-11-16
OBJ-248 Changing create objection endpoint to return 201 status and b…

markpit:
…oth id and status for all created objections inc Ineligibles

Changing the create objection endpoint to not return status 400 when company ineligible as this caused the web app to log errors due to the 400 code indicating that something was wrong with the request.
If an objection is created, a 201 response is returned inc id and status in the body.


markpit:
closing and re-opening to trigger sonar PR analysis.

PR #113, opened 2020-12-10
Obj 451 uri variables when calling file transfer

ICoxCH:
Spring monitors calls to external apis.
By constructing the url from a string each time, each call was counted as a seperate call.
Using URI variables allows spring to recognize each call as the same with different paramaters


SVarallo (approved):
Looks good to me but would suggest a second opinion.

markpit (approved):

PR #116, opened 2021-01-14
BI-6577 Implement sending Chips data via Kafka

markpit:
This now uses the kafka-models project to construct the kafka message for chips-rest-interfaces-consumer.

I've moved AvroSerialiser out from the email package into the common package and added new method to serialise a kafka-model class. this means that we no longer need the ChipsRestInterfacesConsumerContent model class which has now been removed.

The toggling between sending Chips data via rest or kafka is controlled by the ACTIVE_SPRING_PROFILE=chips-rest-interfaces-kafka property. If the value is set as shown, it will use kafka, if it is different then it uses Rest.


src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsKafkaClient.java
```diff
@@ -1,69 +1,112 @@
...
             future.get();
+
+            logger.infoContext(requestId,
+                    "Finished sending kafka message to Chips Rest Interfaces Consumer",
+                    dataForInfoLogMessage);
         } catch (IOException | ExecutionException e) {
             logger.errorContext(requestId, e);
-            return;
-            //throw new ServiceException(e.getMessage());
+            throw new ServiceException(e.getMessage());
```
ICoxCH, line 89: for both of these throw service exceptions, i'd suggest
throw new ServiceException(message, e); so the root cause is still included in the exception

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/email/KafkaEmailClient.java
```diff
@@ -6,6 +6,7 @@
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.strikeoffobjections.common.AvroSerializer;
```
ICoxCH, line 9: this import has been added but nothing in the class has changed, is it needed?

markpit: yeah, the AvroSerializer was in the same package previously so no need to import, it's now moved to common

src/main/java/uk/gov/companieshouse/api/strikeoffobjections/chips/ChipsKafkaClient.java
markpit: yes you're right, I've updated these now and the test

PR #117, opened 2021-01-15
BI-6173 Adding optional sending of Chips data via kafka

markpit:
Added a 'feature flagged' (controlled by a spring profile value - see below) optional method of sending Chips data via Kafka to the Chips Rest Interfaces Consumer app. 
If the config value ACTIVE_SPRING_PROFILE is set to chips-rest-interfaces-kafka then it will send via kafka. If this value is set to anything else, it will send via the current Rest call to Chips Rest Interfaces.


ICoxCH (approved):
Look ok to me

PR #120, opened 2021-01-18
Feature/obj 465 readme

mattch1:
Readme for objections api.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,12 @@ These instructions are for a local docker environment.
 
 ### Endpoints
 
+The full path for each endpoint begins with the app url:
+
+`${API_URL}/company/{COMPANY_NUMBER}/strike-off-objections`
+
+This is common to all the endpoints.
+
 Method    | Path                                                                         | Description
 :---------|:-----------------------------------------------------------------------------|:-----------
 **GET**   | `/eligibility`       | Returns information describing the strike-off eligibility of the company.
--- a/README.md
+++ b/README.md
@@ -1,2 +1,53 @@
 # strike-off-objections-api
+
+### Overview
 API for handling objections to a company being struck off the register
+
+### Requirements
+
+In order to run the service locally you will need the following:
+
+- [Java]
+- [Git](https://git-scm.com/downloads)
+
+### Getting started
+
+To checkout and build the service:
+1. Clone [Docker CHS Development](https://github.com/companieshouse/docker-chs-development) and follow the steps in the README.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,57 @@
...
 API for handling objections to a company being struck off the register
+
+### Requirements
+
+In order to run the service locally you will need the following:
+
+- [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
+- [Maven](https://maven.apache.org/download.cgi)
+- [Git](https://git-scm.com/downloads)
+- [MongoDB](https://www.mongodb.com/)
```
ICoxCH, line 13: You don't need mongoDB installed locally to run the app, docker will spin it up from a pre made container

README.md
```diff
@@ -1,2 +1,57 @@
...
+
+### Getting started
+
+To checkout and build the service:
+1. Clone [Docker CHS Development](https://github.com/companieshouse/docker-chs-development) and follow the steps in the README.
+2. Run ./bin/chs-dev modules enable strike-off-objections
+3. Run ./bin/chs-dev development enable strike-off-objections-api (this will allow you to make changes).
+4. Run docker using "tilt up" in the docker-chs-development directory.
+5. Use spacebar in the command line to open tilt window - wait for strike-off-objections-web to become green.
+6. Open your browser and go to page http://chs.local/strike-off-objections
```
ICoxCH, line 23: As this is the api probably better to have a list of endpoints in the readme, and direct them to hit these. As this url is for the strike-off-api

ICoxCH: As there is now a table of endpoints, i'd change step 6 to something like "Send a rest request to one of the api's endpoints"

README.md
```diff
@@ -1,2 +1,72 @@
...
+7. Work though the pages using example company THE BEAR RETAIL 05916434
+
+These instructions are for a local docker environment.
+
+
+### Endpoints
+
+Method    | Path                                                                         | Description
+:---------|:-----------------------------------------------------------------------------|:-----------
+**GET**   | `/eligibility`       | Returns information describing the strike-off eligibility of the company.
```
ICoxCH, line 32: These paths should be prefixed with ${API_URL}/company/{COMPANY_NUMBER}/strike-off-objections/

mattch1: That would make the paths very long and more difficult to tell each one apart as it is the same for all paths then it would be better to mention this above.

sroberts2 (approved):

PR #131, opened 2021-12-14
Override log4j to safe version and update CH logging dependency to 1.…

mouhajer-ch:
Override log4j transitive dependencies to safe version and update CH logging dependency to 1.9.10

Resolves [DEBT-1412](https://companieshouse.atlassian.net/browse/DEBT-1412)


CHdlewis (approved):
LGTM

PR #140, opened 2022-08-11
Another go at addressing issues in strike off objection api

adr1ancho:


hgurung1 (approved):
Note: Concourse has been pinned so as to not include this version, this is a wip to test whether it addresses the issues in SonarQube

hgurung1 (approved):

PR #156, opened 2024-02-12
Add third lb for internalapi, used by web service

bcullerton:
Add third load balancer route path to strike-off-objections-api:

https://github.com/companieshouse/strike-off-objections-web/blob/master/src/modules/sdk/objections/service.ts#L17


hselvakumar (approved):
Looks good

